PR #71, opened 2020-10-01
Estimate pins backend

atatmaja:
Spec with more details: https://docs.google.com/document/d/18n6-r2kTEKOzVTfhOX0NgKA6xBkix3m8X7Tb7IXhBW4/edit#

Note: This depends on PR #69


app/utils/get_filtered_records.py
```diff
@@ -37,6 +37,13 @@ def get_all_records():
 
         # Alias for country name
         fields_list.append(Country.country_name.label("country"))
+        # Aliases for lat lngs
+        fields_list.append(Country.latitude.label("country_latitude"))
+        fields_list.append(Country.longitude.label("country_longitude"))
+        fields_list.append(State.latitude.label("state_latitude"))
+        fields_list.append(State.longitude.label("state_longitude"))
+        fields_list.append(City.latitude.label("city_latitude"))
+        fields_list.append(City.longitude.label("city_longitude"))
```
joanna-chen, line 46: could we just have one long/lat depending on what estimate grade it is to reduce the amount of logic the frontend needs to do? 
(you could also do this in another PR if you want)

joanna-chen (commented):
I tested some API calls on this branch and it looked good to me, probably have someone more plugged in on the backend side of things to review this too though

app/utils/get_filtered_records.py
atatmaja: Will probably address in another PR

atatmaja:
I think I'll probably have to make some changes to this once the unit tests pr goes in (updating unit tests for affected endpoints to reflect new DB schema and updates to responses). Will take this PR down once unit testing pr goes in, make changes, and make updated PR

app/database_etl/etl_main.py
```diff
@@ -256,6 +243,29 @@ def validate_records(airtable_source):
...
+    data = r.json()
+    coords = None
+    if data is not None and "features" in data and len(data['features']) > 0:
+        # Check if we have any results of the right place type
+        # Otherwise fallback to the first result we get
+        for feature in data['features']:
+            if place_type in feature['place_type']:
+                coords = feature['center']
+                break
+        if coords is None:
```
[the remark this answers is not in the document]
simonarocco: could also set default of coords to be data['features'][0]['center'] and then override it in the for loop as you are currently setting it now. then wouldn't have to do check if coords is not None at the end.

app/database_etl/etl_main.py
```diff
@@ -284,22 +294,46 @@ def main():
     # Create airtable source df
     airtable_source = create_airtable_source_df(data)
 
+    # Create country table df
+    country_df = pd.DataFrame(columns=['country_name', 'country_id'])
+    country_df['country_name'] = airtable_source['country'].unique()
+    country_df['country_id'] = [uuid4() for i in range(len(country_df['country_name']))]
```
simonarocco, line 300: [uuid4() for name in country_df['country_name']]

app/database_etl/etl_main.py
```diff
@@ -256,6 +243,29 @@ def validate_records(airtable_source):
...
+        # Otherwise fallback to the first result we get
+        for feature in data['features']:
+            if place_type in feature['place_type']:
+                coords = feature['center']
+                break
+        if coords is None:
+            coords = data['features'][0]['center']
+    return coords
+
+def add_latlng_to_df(place_type, name_col, df):
```
simonarocco, line 263: maybe we can change the parameter "name_col" to "place_type_name" so it's a bit clearer that it relates to the place

simonarocco (commented):
good. some nit stuff to make more concise. main question is why we aren't adding a country bridge table? seems like by omitting country bridge table we are having to add in more specialized etl code to account for this. i think it would just be easier to add the country bridge table in too so it's consistent with all the other tables that branch off of airtable source.

app/serotracker_sqlalchemy/models.py
```diff
@@ -37,13 +37,23 @@ class AirtableSource(db.Model):
     estimate_grade = Column(String(32))
     created_at = Column(DateTime)
 
+class Country(db.Model):
```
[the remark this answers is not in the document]
atatmaja: because it's a one to many mapping, not many to many

app/utils/get_filtered_records.py
atatmaja: @joanna-chen created the following ticket for this https://airtable.com/tbli2lWQHAqBa6ZcI/viw8Ro4zYPYcBGXRw/reczv1TbukdkC5fEp?blocks=bipIkdtYHmboUfJut

I think I'm going to return an extra field with the pin lat/lng. Don't wanna remove all the other lat/lngs from the response in case we end up needing them for some reason. Lmk if you'd prefer otherwise.

PR #74, opened 2020-10-10
Fix response schema of get_filtered_records when columns argument supplied

atatmaja:
Before:
image

After: 
image


ewanmay (commented):
mmmm yes this does look very useable now

simonarocco (approved):
lgtm sry

PR #75, opened 2020-10-10
Allow empty filters to be passed in

atatmaja:
- Allow filters in the form {filter_key: []} to be passed in for frontend convenience.
- If an empty filter is passed in, it should not be applied (i.e. should let all records pass)


simonarocco (approved):
lgtm

PR #76, opened 2020-10-11
Readd Airtable Scraper Namespace

atatmaja:
Readding legacy airtable_scraper endpoints to ensure backwards compatibility with research


simonarocco (approved):
lgtm. will we ever delete this this or no.

rahularoradfs:
> 
> 
> lgtm. will we ever delete this this or no.

will look at in maybe a month's time, when we have articles submitted - ideally I will integrate my data cleaning code as part of the backend and we will end up with only one db for both research and dashboard tasks. for now, keeping for stability.

PR #81, opened 2020-10-25
Update 'database up to date' date

atatmaja:
https://airtable.com/tbli2lWQHAqBa6ZcI/viw8Ro4zYPYcBGXRw/recpTww5ZJyPlQIp2?blocks=bipIkdtYHmboUfJut 

Update the "server up to date" string that will be shown in the dashboard to be reflective of the last publication date

### DIFF
--- a/app/namespaces/airtable_scraper/airtable_scraper_controller.py
+++ b/app/namespaces/airtable_scraper/airtable_scraper_controller.py
@@ -58,7 +58,7 @@ class AirtableScraper(Resource):
                     max_sampling_end_date = d
         max_sampling_end_date = max_sampling_end_date.strftime('%Y-%m-%d')
 
-        result = {"airtable_request_status_code": status_code, "records": records, "last_publication_date": max_sampling_end_date}
+        result = {"airtable_request_status_code": status_code, "records": records, "updated_at": max_sampling_end_date}
         return jsonify(result)
 
 
--- a/app/namespaces/airtable_scraper/airtable_scraper_controller.py
+++ b/app/namespaces/airtable_scraper/airtable_scraper_controller.py
@@ -49,7 +49,16 @@ class AirtableScraper(Resource):
             # Get all records from using airtable API
             records, status_code = get_all_records(visualize_on_serotracker_filter, airtable_fields_json_name)
             file_created_datetime = datetime.now()
-        result = {"airtable_request_status_code": status_code, "records": records, "updated_at": file_created_datetime}
+
+        max_sampling_end_date = datetime.min
+        for record in records:
+            if record["PUB_DATE"] is not None:
+                d = datetime.strptime(record["PUB_DATE"][0], '%Y-%m-%d')
+                if d > max_sampling_end_date:
+                    max_sampling_end_date = d
+        max_sampling_end_date = max_sampling_end_date.strftime('%Y-%m-%d')
+
+        result = {"airtable_request_status_code": status_code, "records": records, "last_publication_date": max_sampling_end_date}
         return jsonify(result)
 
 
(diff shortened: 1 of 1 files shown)


app/namespaces/airtable_scraper/airtable_scraper_controller.py
```diff
@@ -49,7 +49,16 @@ def get(self):
...
+
+        max_sampling_end_date = datetime.min
+        for record in records:
+            if record["PUB_DATE"] is not None:
+                d = datetime.strptime(record["PUB_DATE"][0], '%Y-%m-%d')
+                if d > max_sampling_end_date:
+                    max_sampling_end_date = d
+        max_sampling_end_date = max_sampling_end_date.strftime('%Y-%m-%d')
+
+        result = {"airtable_request_status_code": status_code, "records": records, "updated_at": max_sampling_end_date}
```
atatmaja, line 61: "updated_at" isn't really an accurate field name anymore, but leaving it in there for backwards compatibility

brett-stephen (approved):

app/namespaces/airtable_scraper/airtable_scraper_controller.py
brett-stephen: alright, understandable

brett-stephen:
Looks good!

app/namespaces/airtable_scraper/airtable_scraper_controller.py
```diff
@@ -49,7 +49,16 @@ def get(self):
             # Get all records from using airtable API
             records, status_code = get_all_records(visualize_on_serotracker_filter, airtable_fields_json_name)
             file_created_datetime = datetime.now()
-        result = {"airtable_request_status_code": status_code, "records": records, "updated_at": file_created_datetime}
+
+        max_sampling_end_date = datetime.min
```
rahularoradfs, line 53: This variable should be max_pub_date throughout - we care about the publication date here, not the study end date!
The field you're accessing reflects that, but the variable name should also

app/namespaces/airtable_scraper/airtable_scraper_controller.py
```diff
@@ -49,7 +49,16 @@ def get(self):
...
+        for record in records:
```
rahularoradfs, line 54: Potentially a more elegant solution for this for loop (will need testing):
`record["PUB_DATE"].dropna().explode().to_datetime(format '%Y-%m-%d').max()`

dropna() takes care of the is not None
explode() expands each list-like item as its own row, meaning you don't need to index, and are now actually taking all pub dates into account 
to_datetime is the pandas builtin for strptime

rahularoradfs (changes requested):
per comments - variable naming and suggested more pandas-like way to refactor for loop!

app/namespaces/airtable_scraper/airtable_scraper_controller.py
atatmaja: lol whoops forgot to update that

app/namespaces/airtable_scraper/airtable_scraper_controller.py
atatmaja: don't think we're dealing with a df here, and not sure if it's worth it to convert to a df and back to a python dict (b/c we need to send to client in that format) just to get this marginal improvement

@simonarocco what do you think

simonarocco: i think readability of austin's solution makes it a bit more desirable. for loops can be optimized when they're complex however this is quite short, and i think going forward it's fine to keep short for loops like this in just bc it will make it easier to read through quick and understand what's going on without having to look up certain built in df/series funcs.

simonarocco (approved):

app/namespaces/airtable_scraper/airtable_scraper_controller.py
rahularoradfs: good point -- have been so buried in stats these past two weeks that I saw `data_struct['index']` and immediately assumed df. Would be in favor of this solution if it was, but definitely not worth converting back and forth.

rahularoradfs (approved):

PR #83, opened 2020-10-29
Add filter options endpoint

atatmaja:


app/utils/get_filter_options.py
```diff
@@ -0,0 +1,50 @@
...
+from sqlalchemy import distinct, func
+from itertools import groupby
+from functools import reduce
+
+def get_all_filter_options():
+    with db_session() as session:
+        # List of fields in AirtableSource to query
+        airtable_field_strings = ['source_type', 'study_status', 'overall_risk_of_bias', 'sex', 'estimate_grade']
+        options = {}
+        for field_string in airtable_field_strings:
```
simonarocco, line 14: why don't you just combine this for loop and the one below into one and add the airtable_field_strings to the other_fields dict and just set their table to AirtableSource? not sure if there is a need to have them separate

simonarocco (commented):
lgtm just see my one suggestion

app/utils/get_filter_options.py
atatmaja: yeah fair point

atatmaja: actually I'm going to push back on this. for all the other_fields, there is only one field per table that we need to scrape, and that field's name must be changed. that's a bit different from the fields we need to scrape from airtable_source. 

could probably refactor into one loop but I think this is pretty readable and will not really incur much of a perf diff.

PR #84, opened 2020-10-31
Changes for frontend rearch

ewanmay:
- Time string changes
- Extra columns
- Filtering issues


.gitignore
```diff
@@ -18,3 +18,5 @@ __pycache__/
 
 # Virtualenv
 venv/
+
+serotracker/
```
atatmaja, line 22: what's this for lol

app/namespaces/meta_analysis/meta_analysis_controller.py
```diff
@@ -54,7 +54,8 @@ def post(self):
         # Query all the records with the desired filters. Pull only country, denom, and seroprev cols
         filters = json_input.get('filters', None)
         start_date, end_date = convert_start_end_dates(json_input)
-        columns = ['country', 'denominator_value', 'serum_pos_prevalence']
+        columns = ['country', 'denominator_value', 'sex', 'age', 'study_status', 'population_group',
```
atatmaja, line 57: we always need denominator_value and serum_pos_prevalence but not the other columns (they depend on the aggregation var the user provides). can we parameterize that instead of always getting all of the columns?

atatmaja (commented):
looks good, 1 minor suggestion

app/namespaces/meta_analysis/meta_analysis_controller.py
ewanmay: I had to add these columns in order to change the y axis aggregation variable on the charts. If there's a better way then let's do that

.gitignore
ewanmay: yeah rip my venv

atatmaja (approved):
LGTM

PR #91, opened 2020-11-11
Fix bug where meta analysis endpoint was failing when aggregating by a multi select field e.g. sex, age, pop group

atatmaja:


rahularoradfs:
Solution seems eminently reasonable to me, and `explode` behaves well even with single-select fields and shouldn't raise any issues there. Will leave a proper review to folks who are closer to the code, but it's a really nice catch!

rahularoradfs (commented):
Solution seems eminently reasonable to me, and `explode` behaves well even with single-select fields and shouldn't raise any issues there. Will leave a proper review to folks who are closer to the code, but it's a really nice catch!

simonarocco (approved):
lgtm

PR #95, opened 2020-11-22
Fix bug where records were being retrieved in cases where we should be retrieving no records

atatmaja:


joanna-chen (approved):
looks good

PR #97, opened 2020-11-24
add hyperlocal as estimate grade

simonarocco:
add hyperlocal as estimate grade option for returning country seroprev summary. before this was not an option so filtering by hyperlocal was incorrect.


atatmaja (approved):
LGTM

PR #100, opened 2020-11-28
Data cleaning PR #1

simonarocco:
…emoved in old airatble scraper code and ETL)

- deleted study status column, and renamed included_n column to include
- deleted geography type column from research options


atatmaja (approved):
LGTM

PR #101, opened 2020-11-29
Data cleaning #2

simonarocco:
- adding filtering to remove rows if included, serum pos prev, denominator or sampling end are null
- changed to grab first author last name based on most recent pub date
- deleted following bridge tables: approving regulator, age, test type, specimen type and added these to airtable source instead
- changing Not Reported, Not reported, and NR to None
- take best risk of bias and apply to all estimates in study
- deleted population group bridge table
- converted city, state and test manufacturer from comma sep string to list


app/database_etl/etl_main.py
```diff
@@ -205,10 +203,8 @@ def load_postgres_tables(tables_dict, engine):
 
 
 def drop_old_entries():
-    all_tables = [AirtableSource, City, State, Age, PopulationGroup,
-                  TestManufacturer, ApprovingRegulator, TestType, SpecimenType, Country,
-                  CityBridge, StateBridge, AgeBridge, PopulationGroupBridge,
-                  TestManufacturerBridge, ApprovingRegulatorBridge, TestTypeBridge, SpecimenTypeBridge]
+    all_tables = [AirtableSource, City, State, TestManufacturer, Country,
```
atatmaja, line 206: wow this looks like a pretty small list of tables now!!

app/database_etl/etl_main.py
```diff
@@ -254,13 +251,63 @@ def get_coords(place_name, place_type):
...
+
 def add_latlng_to_df(place_type, place_type_name, df):
     df['coords'] = df[place_type_name].map(lambda a: get_coords(a, place_type))
     df['longitude'] = df['coords'].map(lambda a: a[0] if isinstance(a, list) else None)
     df['latitude'] = df['coords'].map(lambda a: a[1] if isinstance(a, list) else None)
     df = df.drop(columns=['coords'])
     return df
 
+
+def get_most_recent_pub_date_info(row):
```
atatmaja, line 263: nit: maybe we can call this `get_most_recent_publication_info` because the function is scraping more than just the publication_date, but all the publication info at large

app/database_etl/etl_main.py
```diff
@@ -254,13 +251,63 @@ def get_coords(place_name, place_type):
...
+        row['source_type'] = row['source_type'][max_index]
+
+    # Index whether org author exists and corresponding first author
+    is_org_author = row['organizational_author'][max_index]
+    row['organizational_author'] = is_org_author
+    row['first_author'] = row['first_author'][max_index]
+
+    # If it is not an organizational author, then get last name
+    if not is_org_author:
+        row['first_author'] = row['first_author'].strip().split()[-1]
```
atatmaja, line 285: Need to make sure that .strip().split()[-1] works with an empty string. I think that it would give you index out of range.

app/database_etl/etl_main.py
```diff
@@ -254,13 +251,63 @@ def get_coords(place_name, place_type):
...
+    return row
+
+
+def apply_min_risk_of_bias(df):
+    bias_hierarchy = ['Low', 'Moderate', 'High', 'Unclear']
+    for name, subset in df.groupby('study_name'):
+        if (subset['overall_risk_of_bias']).isnull().all():
+            subset['overall_risk_of_bias'] = 'Unclear'
+            continue
+        for level in bias_hierarchy:
```
atatmaja, line 295: nit: I think we can simplify this logic here by just having our bias_hierarchy consist of ['Low', 'Moderate', 'High'], that way we wouldn't need that extra check for level == 'Unclear'

app/database_etl/etl_main.py
```diff
@@ -254,13 +251,63 @@ def get_coords(place_name, place_type):
...
+                continue
+    return df
+
+
+def get_city(row):
+    if row['city']:
+        return row['city'].split(',')
+    elif row['county']:
+        return row['county'].split(',')
+    else:
```
atatmaja, line 307: nit: So this else case would happen if there's no city or county right? Would it be better to just return None or '' instead of returning row['city'], seems slightly confusing

app/database_etl/etl_main.py
```diff
@@ -273,25 +320,41 @@ def main():
...
-                         'test_type', 'specimen_type']
+    single_element_list_cols = ['included', 'source_name', 'url', 'source_publisher', 'summary',
+                                'study_type', 'country', 'lead_organization', 'overall_risk_of_bias']
 
     # Remove lists from single select columns
     for col in single_element_list_cols:
         data[col] = data[col].apply(lambda x: x[0] if x is not None else x)
 
+    # Convert elements that are "Not reported" or "Not Reported" or "NR" to None
+    data.replace({'NR': None, 'Not Reported': None, 'Not reported': None}, inplace=True)
```
atatmaja, line 331: lmao

atatmaja (commented):
Couple nits: Good progress!

PR #119, opened 2020-12-20
Estimate pins improvements

atatmaja:
Changelog:
- Leverage ISO 3166 country alpha2codes when performing mapbox geocoding queries for better accuracy (see https://docs.mapbox.com/api/search/geocoding/#mapboxplaces-permanent)
- Average the locations of regions or cities if multiple exist for a given record
- Implement "pin jittering" to ensure that pins at the same location don't overlap exactly (for visualization purposes)


app/database_etl/etl_main.py
```diff
@@ -34,6 +34,8 @@ def read_from_json(path_to_json):
 dirname = os.path.dirname(__file__)
 filename = os.path.join(dirname, 'country_iso3.json')
 ISO3_CODES = read_from_json(filename)
+filename = os.path.join(dirname, 'country_iso2.json')
```
simonarocco, line 37: can we refactor these lines? make them a function where u pass in the file name or smth so it's more organized?

app/database_etl/etl_main.py
```diff
@@ -348,22 +359,38 @@ def apply_study_max_estimate_grade(df):
...
     else:
         return row['city']
 
 
+# Returns "state_countryCode"
+# Needed becuase country code is used to limit mapbox API
+# geosearch queries (improving query accuracy)
+def add_country_code_to_state(row):
+    if row['state']:
+        return [f"{state}_{get_country_code(row['country'], iso3=False)}"
```
simonarocco, line 388: can you just make this a for loop and return a variable? it's a bit long of a list comp

simonarocco (commented):
minor nits

app/database_etl/etl_main.py
atatmaja: sure thing

app/database_etl/etl_main.py
atatmaja: sure

PR #123, opened 2021-01-16
Research source endpoint

simonarocco:
- added a research_fields boolean param to /records get and post endpoints
- in the get endpoint, this is a request param, in the post endpoint, this is part of the json body
- if set to True, all columns from research source table will be returned as well
- same filtering as before

### DIFF
--- a/app/utils/get_filtered_records.py
+++ b/app/utils/get_filtered_records.py
@@ -1,20 +1,22 @@
 from itertools import groupby
 from functools import reduce
-from app.serotracker_sqlalchemy import db_session, DashboardSource, db_model_config, Country, State, City
+from app.serotracker_sqlalchemy import db_session, DashboardSource, ResearchSource, \
+    db_model_config, Country, State, City
 import pandas as pd
 import numpy as np
 from .estimate_prioritization import get_prioritized_estimates
 from statistics import mean
 
 
-def get_all_records():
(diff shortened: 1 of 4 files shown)


app/utils/get_filtered_records.py
```diff
@@ -23,10 +25,26 @@ def get_all_records():
...
                          'test_adj', 'specimen_type', 'test_type', 'population_group', 'url']
 
+        # Add columns from dashboard source to select statement
         fields_list = [DashboardSource.source_id]
         for field_string in field_strings:
             fields_list.append(getattr(DashboardSource, field_string))
 
+        # If research fields is True, add columns from research source to select statement
+        if research_fields:
+            research_source_cols = ['case_population', 'deaths_population', 'age_max', 'age_min', 'age_variation',
```
atatmaja, line 35: do we want to put this in some config?

simonarocco: i didnt put it in a config because the list of fields for dashboard source above was not in a config

atatmaja (approved):
Looks fine as long as it's been tested. One minor comment

app/utils/get_filtered_records.py
atatmaja: okay leaving it up to your judgement

PR #136, opened 2021-01-26
Add logging to meta analysis namespace, suppress stream logging

atatmaja:
# Changelog
- Change StreamHandler level to "ERROR" to ensure only error messages are outputted to the console
- Add logging statements to the meta analysis endpoint

# How did I test?
- Ran the server locally and pointed a local version of the ST web client to it
- Navigated to the analyze tab and changed some filters
- Checked the logfile to ensure that the meta-analysis endpoint invocations were properly logged


simonarocco (approved):
lgtm

PR #137, opened 2021-01-26
Fix param parsing for research_fields and prioritize_estimates

simonarocco:
Previously we were setting the defaults of the research_fields and prioritize_estimates fields to be False and True respectively, and we set the type to be bool. But in python bool(True) = bool(False) = True because the value is set, so if you passed in research_fields = False it would still set the param to true (reverse logic for prioritize_estimates). I changed the param type to be a string and then manually set the bool val of the param based on the value of the string.

### DIFF
--- a/app/namespaces/data_provider/data_provider_controller.py
+++ b/app/namespaces/data_provider/data_provider_controller.py
@@ -21,8 +21,8 @@ class Records(Resource):
         reverse = request.args.get('reverse', None, type=bool)
         page_index = request.args.get('page_index', None, type=int)
         per_page = request.args.get('per_page', None, type=int)
-        research_fields = request.args.get('research_fields', False, type=bool)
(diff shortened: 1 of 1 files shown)


atatmaja (approved):
Hmm interesting, maybe this contributes to why our site was loading slowly...

Good catch, can you just add a comment about the problem right above those lines you changed so that we aren't confused in the future?

rahularoradfs:
This is so strange!! Could definitely see this being a contributor to slow loads.

A couple of potentially related threads, seems like GET params can only be strings, but I'm sure you've got this figured already Simona -- 
- [this GitHub thread](https://github.com/flask-restful/flask-restful/issues/397) 
- [StackOverflow](https://stackoverflow.com/questions/29739970/how-to-pass-boolean-variable-of-false-in-a-url)

PR #139, opened 2021-01-27
Revert logging to use default flask logging

atatmaja:
We have a strong hypothesis that logging has caused high CPU usage on our EC2 instance and therefore sluggish behaviour. This PR reverts the custom logging work that we've done temporarily. Once this PR is merged, we should be using default flask logging again.

Plan moving forward: monitor CPU usage and request speed on prod. If these return back to normal, then we should experiment with our custom monitoring solutions on a temporary EC2 instance to see if there's anything we can do to improve logging performance while maintaining utility.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -7,9 +7,9 @@ from flask_cors import CORS
 from flask_sqlalchemy import SQLAlchemy
 from .config import config_by_name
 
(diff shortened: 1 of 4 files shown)


simonarocco (approved):
lgtm

simonarocco (changes requested):
make easier :)

PR #141, opened 2021-01-28
ETL Patch

atatmaja:
### Changelog
- If a table fails while it is being inserted into postgres, drop all records from that table from the current run (i.e. just leave the table as is from the previous ETL run)
- TODO: add slack message that logs the error
- Stripping strings of \x00 characters (NULL characters in strings)

### DIFF
--- a/app/database_etl/etl_main.py
+++ b/app/database_etl/etl_main.py
@@ -235,11 +235,11 @@ def load_postgres_tables(tables_dict, engine):
     return True
 
 
-def drop_table_entries(type='old'):
+def drop_table_entries(drop_old=True):
     for table_name in table_names_dict:
         table = table_names_dict[table_name]
         with db_session() as session:
-            if type == 'old':
+            if drop_old:
                 # Drop old records if type is old
                 session.query(table).filter(table.created_at != CURR_TIME).delete()
             else:
@@ -653,10 +653,10 @@ def main():
 
     # If all tables were successfully loaded, drop old entries
     if load_status:
-        drop_table_entries(type='old')
+        drop_table_entries(drop_old=True)
     # Otherwise drop entries from current ETL run
     else:
-        drop_table_entries(type='new')
+        drop_table_entries(drop_old=False)
 
     # Make sure that filter options are still valid
     check_filter_options(dashboard_source)
--- a/app/database_etl/etl_main.py
+++ b/app/database_etl/etl_main.py
@@ -226,41 +226,26 @@ def load_postgres_tables(tables_dict, engine):
                                con=engine,
                                if_exists='append',
                                index=False)
-            drop_old_entries(table_name)
         except (SQLAlchemyError, ValueError) as e:
             # Send error email
             logging.error(e)
             # TODO: send slack message with error (include the table name that failed)
+            # Break out of function and return 'failure'
+            return 'failure'
+    return 'success'
 
-            # Delete  records in table that failed which contain current datetime (remove records from current ETL run)
-            with db_session() as session:
-                # Drop record if it was not added during the current run
-                table = table_names_dict[table_name]
+
+def drop_table_entries(type='old'):
+    for table_name in table_names_dict:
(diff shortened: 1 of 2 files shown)


app/database_etl/etl_main.py
```diff
@@ -461,7 +474,7 @@ def check_filter_options(dashboard_source):
         # Check to see if the new options are equal to the curr hardcoded options
         if new_options != set(curr_filter_options[filter_type]):
             changed_filter_options[filter_type] = new_options
-            logger.info(new_options)
+            logging.info(new_options)
```
abeljoseph, line 477: Should this be `logging.info` or `logger.info`?

abeljoseph (approved):

app/database_etl/etl_main.py
simonarocco: ya it should be logging

app/database_etl/etl_main.py
```diff
@@ -203,21 +220,49 @@ def create_bridge_tables(original_data, multi_select_tables):
 def load_postgres_tables(tables_dict, engine):
     # Load dataframes into postgres tables
     for table_name, table_value in tables_dict.items():
-        table_value.to_sql(table_name,
-                           schema='public',
-                           con=engine,
-                           if_exists='append',
-                           index=False)
+        if table_name == 'research_source':
+            table_value.loc[table_value['estimate_name'] == '0724_DeKalbFulton_CDCCOVID-19ResponseTeam_Female_PopAdj',
```
atatmaja, line 224: what's the purpose of this lol

atatmaja: if this is just test code pls remove

simonarocco: OH LMAO just test code to F**CCKK it up haha

app/database_etl/etl_main.py
```diff
@@ -203,21 +220,49 @@ def create_bridge_tables(original_data, multi_select_tables):
...
+                               con=engine,
+                               if_exists='append',
+                               index=False)
+            drop_old_entries(table_name)
+        except (SQLAlchemyError, ValueError) as e:
+            # Send error email
+            logging.error(e)
+            # TODO: send slack message with error
+
+            # Delete  records in table that failed which contain current datetime (remove records from current ETL run)
```
atatmaja, line 235: hmm should we just fully revert to the pre etl state if we see a failure, and not just revert the current table?

reason I'm suggesting this is because if table A is reverted but not table B, then the bridge table between table A and B might be fucked up

simonarocco: mmm that is a very good point

atatmaja (commented):
Thanks for this, just had one question/suggestion

app/database_etl/etl_main.py
```diff
@@ -203,22 +220,32 @@ def create_bridge_tables(original_data, multi_select_tables):
...
+                               schema='public',
+                               con=engine,
+                               if_exists='append',
+                               index=False)
+        except (SQLAlchemyError, ValueError) as e:
+            # Send error email
+            logging.error(e)
+            # TODO: send slack message with error (include the table name that failed)
+            # Break out of function and return 'failure'
+            return 'failure'
```
atatmaja, line 234: nit: True/False vs 'success' and 'failure'?

simonarocco: sure

app/database_etl/etl_main.py
```diff
@@ -203,22 +220,32 @@ def create_bridge_tables(original_data, multi_select_tables):
...
+    return 'success'
+
+
+def drop_table_entries(type='old'):
```
atatmaja, line 238: nit: same here, boolean arg vs string arg here

simonarocco: ok

atatmaja (commented):
I can't approve my own pr apparently but Approved, just some minor nits

atatmaja:
lgtm

PR #142, opened 2021-01-28
Fix get record details endpoint

atatmaja:
Changelog:

- Use common list of "dashboard_source_cols" across the record_details and record endpoints
- Ensure that the returned record is always a dict rather than an array of length 1
- Convert dates to isoformat before returning
- Ignore None values in multi select fields

### DIFF
--- a/app/namespaces/data_provider/data_provider_service.py
+++ b/app/namespaces/data_provider/data_provider_service.py
@@ -3,7 +3,7 @@ import numpy
 import math
 from random import uniform
 
-from app.serotracker_sqlalchemy import db_session, DashboardSource, Country, db_model_config
+from app.serotracker_sqlalchemy import db_session, DashboardSource, Country, db_model_config, dashboard_source_cols
 from sqlalchemy import case, and_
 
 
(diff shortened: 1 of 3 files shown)


app/namespaces/data_provider/data_provider_service.py
```diff
@@ -46,7 +46,7 @@ def _get_parsed_record(results):
     for col in results_df.columns:
         # Convert multi select values into an array
         if col in multi_select_cols:
-            multi_select_vals = list(set(getattr(results_df, col).tolist()))
+            multi_select_vals = [e for e in list(set(getattr(results_df, col).tolist())) if e is not None]
```
simonarocco, line 49: can just say "if e" instead of "if e is not None"

simonarocco (commented):
lgtm, nit

app/namespaces/data_provider/data_provider_service.py
atatmaja: Hmm don't think this should ever happen but this suggestion would behave unexpectedly if one of our multi select vals is the integer 0. 

Imo if e is not None is not too unreadable and is more explicit about what we're trying to guard against.

simonarocco: sg

PR #144, opened 2021-01-31
Changing all error logging emails to slack notifications

simonarocco:
## Describe the feature or bug this PR addresses.
- Changing all email notifications in app and ETL to be slack notifications

- I triggered errors in the ETL and app and looked at the output notifications in the slack channels

- No, but on a related note, @abeljoseph you need to change the Slack channel names in your scripts
- @abeljoseph you will have to add an environment variable called 'SLACKBOT_TOKEN'


app/utils/notifications_sender.py
```diff
@@ -0,0 +1,49 @@
...
+        body += f"Record: {record}\n"
+        body += f"Errors: {errors}\n\n"
+
+    body += "\nSincerely,\nIIT Backend Alerts"
+    send_slack_message(body, channel='#dev-logging-etl')
+    return
+
+
+def send_slack_message(message, channel='#dev-logging-backend'):
+    bot_token = os.getenv('BOT_TOKEN')
```
atatmaja, line 46: nit: Can we call this SLACKBOT_TOKEN, so that we don't get confused in the future?

atatmaja (approved):
Minor comment about env var naming, but lgtm

PR #146, opened 2021-01-31
re-enabling logging

simonarocco:
- Re-enable logging
- All INFO and ERROR logs will be written to logfile.log
- All ERROR logs will be printed to console

- I generated info logs by hitting our data provider endpoints and generated errors and saw that they were logged to the logfile

- @abeljoseph wrote a script to print these logs from logfile to the data logging channel
- The logfile.log will be purged every 30 mins when the script is run


atatmaja:
Code looks fine but did we test this on staging? Maybe we can simulate hitting the endpoint like once every 30 seconds and seeing if CPU usage ever spikes to unreasonable levels.

atatmaja (approved):
LGTM but please monitor the CPU throughout the day after this is merged

PR #147, opened 2021-02-01
Add subgroup specific cat clean col

simonarocco:
- adds a column called subgroup_specific_category which is needed for displaying granular canadian data
- also removes any string length limits on columns in research source

- ran migrations to make sure string limitations are gone in postgres
- ran ETL to make sure new column was added

- run migrations


atatmaja (approved):
LGTM

atatmaja:
NVM, we forgot to commit the migration lol

PR #148, opened 2021-02-02
etl MEGA refactor

simonarocco:
- long overdue major makeover of our ETL
- 1 file -> many files
- much more legible and comprehensive

- ran ETL and made sure it works

- no, no migrations or anything to run


app/database_etl/postgres_tables_handler/table_generator.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def isotype_col(isotype_string, x):
+    # Function for creating isotype boolean columns
+    if x:
+        return True if isotype_string in x else False
+    return False
+
+
+def create_dashboard_source_df(original_data, current_time):
```
atatmaja, line 20: hmm consider making this into a class so that we can reference self.current_time instead of having to add it as an arg to every function call

app/database_etl/postgres_tables_handler/postgres_loader.py
```diff
@@ -0,0 +1,82 @@
+import logging
+
+import pandas as pd
+from marshmallow import ValidationError, INCLUDE
+from sqlalchemy.exc import SQLAlchemyError
+
+from app.utils.notifications_sender import send_schema_validation_slack_notif, send_slack_message
+from app.serotracker_sqlalchemy import db_session, DashboardSource, ResearchSource, Country, City, State, \
+    TestManufacturer, AntibodyTarget, CityBridge, StateBridge, TestManufacturerBridge, AntibodyTargetBridge
+#from ..etl_main import CURR_TIME
```
atatmaja, line 10: remove

app/database_etl/postgres_tables_handler/table_formatter.py
```diff
@@ -0,0 +1,87 @@
...
+    if type(x) != str:
+        return x
+    encoded_val = x.encode('utf-8')
+    if b'\x00' in encoded_val:
+        encoded_val = encoded_val.replace(b'\x00', b'\x20')
+        return encoded_val.decode('utf-8')
+    return x
+
+
+def add_mapped_variables(df):
```
atatmaja, line 17: personally this function seems more related to the functions in `airtable_records_formatter` because it's applying transformations to the df in order to add/update columns

atatmaja (commented):
Couple of suggestions. I also would like to run this locally and make sure that I can get my local web client pointing to the database/working well before we merge this

app/database_etl/airtable_records_handler/airtable_records_formatter.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # df state, city and test_manufacturer fields to lists
+    df['test_manufacturer'] = df['test_manufacturer'].apply(lambda x: x.split(',') if x else x)
+    df['state'] = df['state'].apply(lambda x: x.split(',') if x else x)
+    df['city'] = df.apply(lambda row: get_city(row), axis=1)
+    df['state'] = df.apply(lambda row: add_country_code_to_state(row), axis=1)
+    return df
+
+
+def apply_min_risk_of_bias(df):
```
atatmaja, line 72: similarly, could make a class here to make sure we're not passing df all over the place

app/database_etl/postgres_tables_handler/table_generator.py
simonarocco: making the dashboard source as a class or making a whole class for etl?

app/database_etl/postgres_tables_handler/table_formatter.py
simonarocco: ya, i see what you mean. i put it here because the order of which modules are used in sequential order is: airtable records to just get the records from airtable, then table handler to actually format those tables (which is where the mapped vars function comes in) and then postgres loader.

simonarocco:
> Couple of suggestions. I also would like to run this locally and make sure that I can get my local web client pointing to the database/working well before we merge this

you and @abeljoseph should both do this

app/database_etl/airtable_records_handler/airtable_records_formatter.py
simonarocco: I'm not sure which part you want to make a class

atatmaja (approved):

abeljoseph (approved):
Wonderful!

PR #150, opened 2021-02-04
WIP: Who gbd filter

atatmaja:

Adds a filter for WHO GBD regions. This task involved the following changes:
- Moving the `gbd_region` column from the `research_source` table to the `dashboard_source` table
- Enabling filtering by the `gbd_region` column
- Consolidating lists of all of our table column names and filter names into a single config file so that we only need to update one file the next time we pull in columns or add new filters

- https://airtable.com/tbli2lWQHAqBa6ZcI/viw8Ro4zYPYcBGXRw/recut4lmRL26Mr9Uj 

Haven't thoroughly yet, will convert this from a draft to a regular PR once that is done.

No.


app/utils/filter_option_utils.py
```diff
@@ -6,7 +6,75 @@
...
+        'Central Europe, Eastern Europe, and Central Asia',
+        'High-income',
+        'Latin America and Caribbean',
+        'North Africa and Middle East',
+        'South Asia',
+        'Southeast Asia, East Asia, and Oceania',
+        'Sub-Saharan Africa'
+    ]
+}
 def get_filter_static_options():
```
simonarocco, line 78: This function is copying the code from the above dict directly. Seems like this function isn't really needed, and we can just get the keys from static_filter_options to come up with all_filter_types. If we need this function for some reason because it is used in another file, then have it return static_filter_options. In that case tho, the other file could just import static_filter_options instead of the function that returns it.

app/serotracker_sqlalchemy/db_model_config.py
```diff
@@ -27,19 +27,22 @@
...
 
-dashboard_source_cols = ['source_name', 'source_type', 'study_name', 'denominator_value',
+dashboard_source_cols = ['source_id', 'source_name', 'source_type', 'study_name', 'denominator_value',
                          'overall_risk_of_bias', 'serum_pos_prevalence', 'isotype_igm', 'isotype_iga',
                          'isotype_igg', 'sex', 'age', 'sampling_start_date', 'sampling_end_date', 'estimate_grade',
                          'isotype_comb', 'academic_primary_estimate', 'dashboard_primary_estimate', 'pop_adj',
-                         'test_adj', 'specimen_type', 'test_type', 'population_group', 'url']
+                         'test_adj', 'specimen_type', 'test_type', 'population_group', 'url', 'gbd_region']
 
 research_source_cols = ['case_population', 'deaths_population', 'age_max', 'age_min', 'age_variation',
```
simonarocco, line 47: Seems that research_source_cols is defined in this file and in airtable_fields_config - can we just have it in one place so there is less for us to keep track of and maintain?

simonarocco (commented):
Looks good. A few suggestions. This is a pretty big change, so we need to test thoroughly to make sure that all the cols in dashboard source and research source are getting filled as expected.

app/utils/filter_option_utils.py
atatmaja: Actually the function is preferred over the dict, I forgot to get rid of the dict. 

Would pref the function because the output of the`get_filter_static_options()` function can be modified (which the `get_all_filter_options()` function needs to do). If we were to just use the dict, we would need to make a copy of it in order to modify it.

app/serotracker_sqlalchemy/db_model_config.py
atatmaja: Actually there might be some subtle differences between the airtable_fields_config and db_model_config versions. I'm gonna clean things, close up this PR, and re request review later

atatmaja:
Closing to test and rethink how all of our field configs work

PR #152, opened 2021-02-07
Meta analysis empty records fix

simonarocco:
- Meta analysis was throwing key errors when trying to aggregate records with the aggregation var if there were no records to aggregate
- No records occur if filters remove all records (e.g. source type = Publication is not a valid source type)

- Hit meta analysis endpoint using body in dev-logging-backend channel and see that empty dict is returned

- No but I realized that when we copy the request body from the dev-logging-backend channel it is not in JSON format because there are single quotes instead of double, so you have to modify to change to JSON format before inputting in postman
- @abeljoseph can you modify the script so that if the request body is "m", instead of adding just "m" to the error you add m = json.dumps(m)?

### DIFF
--- a/app/namespaces/meta_analysis/meta_analysis_controller.py
+++ b/app/namespaces/meta_analysis/meta_analysis_controller.py
@@ -64,6 +64,9 @@ class MetaAnalysis(Resource):
         columns = ['country', 'denominator_value', 'serum_pos_prevalence']
         columns.append(agg_var)
         records = get_filtered_records(filters=filters, columns=columns, start_date=start_date, end_date=end_date)
+        if not records:
+            logging.error('No records with specified filters found.')
+            return dict(records)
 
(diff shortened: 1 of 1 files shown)


app/namespaces/meta_analysis/meta_analysis_controller.py
```diff
@@ -64,6 +64,9 @@ def post(self):
         columns = ['country', 'denominator_value', 'serum_pos_prevalence']
         columns.append(agg_var)
         records = get_filtered_records(filters=filters, columns=columns, start_date=start_date, end_date=end_date)
+        if not records:
+            logging.error('No records with specified filters found.')
+            return dict(records)
```
atatmaja, line 69: nit: I would just return dict(), I know that they're functionally the same but it's clearer that we're returning an empty dict that way.

atatmaja (commented):
one nit

PR #154, opened 2021-02-12
Research data processing

simonarocco:
- Send additional columns through the endpoint so they are shown on dashboard
- Normalize ind_se and ind_sp instead of ind_se_n and ind_sp_n
- Add error handling if no gbd region or subregion for country
- Add filtering by publication date (this required changing how we are filtering for sampling start/end date)

- Tested the records endpoint and ETL
- Tested filtering for sampling and publication dates

- No but we need to change how the frontend is filtering for dates because start_date --> sampling_start_date and end_date --> sampling_end_date for filtering for dates in the json payload


app/database_etl/postgres_tables_handler/table_formatter.py
```diff
@@ -29,6 +32,11 @@ def add_mapped_variables(df):
         else:
             gbd_region_col.append(None)
             gbd_subregion_col.append(None)
+
+            # Send slack message and log warning
```
rahularoradfs, line 36: Good catch here - are there other mappings that we could do something like this for? It'd be great to have this in place for the other mappings in the ETL e.g., LMIC/HIC, general population / special populations, and so on.

app/utils/get_filtered_records.py
```diff
@@ -19,11 +19,13 @@ def get_all_records(research_fields=False):
                          'state_longitude', 'state_latitude']
 
         # Create list of fields in DashboardSource to query unless the specific columns are specified
-        field_strings = ['source_name', 'source_type', 'study_name', 'denominator_value',
-                         'overall_risk_of_bias', 'serum_pos_prevalence', 'isotype_igm', 'isotype_iga',
-                         'isotype_igg', 'sex', 'age', 'sampling_start_date', 'sampling_end_date', 'estimate_grade',
-                         'isotype_comb', 'academic_primary_estimate', 'dashboard_primary_estimate', 'pop_adj',
-                         'test_adj', 'specimen_type', 'test_type', 'population_group', 'url']
+        field_strings = ['source_name', 'source_type', 'study_name', 'denominator_value', 'overall_risk_of_bias',
```
rahularoradfs, line 22: Thanks!! Did we get filtering by publication date in here,o or will that be a sep PR?

rahularoradfs (commented):
The other question I've got for you, per Slack, is what filters we're applying to the data coming in. There's no visualize on SeroTracker filter anymore, right? We are ensuring that the Included boolean is true? Do you have checks in there for superceded studies or any other inclusion checks?

app/utils/get_filtered_records.py
atatmaja, line 22: Can leave like this for this PR but @simonarocco wanna sync for like 15-20 sometime around put stuff like this into config files and better organize our existing config files?

simonarocco:
> The other question I've got for you, per Slack, is what filters we're applying to the data coming in. There's no visualize on SeroTracker filter anymore, right? We are ensuring that the Included boolean is true? Do you have checks in there for superceded studies or any other inclusion checks?

@rahularoradfs we actually have no filters that we're applying to the data coming in. We don't have the included boolean is true (we removed this for some reason I'm pretty sure)

atatmaja:
@simonarocco can we make this a draft until its ready for review again

app/utils/get_filtered_records.py
simonarocco: see newest commit

app/database_etl/airtable_records_handler/airtable_records_formatter.py
```diff
@@ -57,14 +57,14 @@ def standardize_airtable_data(df):
 
     # Convert elements that are "Not reported" or "Not Reported" or "NR" to None
     df.replace({'nr': None, 'NR': None, 'Not Reported': None, 'Not reported': None, 'Not available': None},
-                 inplace=True)
+               inplace=True)
 
     # Replace columns that should be floats with NaN from None and rescale to percentage
-    df[['ind_sp_n', 'ind_se_n']] = df[['ind_sp_n', 'ind_se_n']].replace({None: np.nan}) / 100
+    df[['ind_sp', 'ind_se']] = df[['ind_sp', 'ind_se']].replace({None: np.nan}) / 100
```
atatmaja, line 63: why are these integers in airtable given that "sensitivity" and "specificity" are entered in as floats? seems confusing @rahularoradfs 

Would be cleaner to have both of them treated the same way by our code imo

app/namespaces/data_provider/data_provider_schema.py
```diff
@@ -32,8 +32,11 @@ class RecordsSchema(Schema):
                                                                  "estimate_grade", "isotypes_reported", "created_at",
                                                                  "pin_longitude", "pin_latitude", "pin_region_type"])))
     # Date fields should be supplied as unix timestamp
-    start_date = fields.String()
-    end_date = fields.String()
+    sampling_start_date = fields.String()
```
atatmaja, line 35: this name change might break the front end so we need to coordinate pushes

atatmaja: I do agree with the name change tho

app/utils/get_filtered_records.py
```diff
@@ -159,23 +160,38 @@ def should_include(d, k, v):
     else:
         result = query_dicts
 
-    def date_filter(record, start_date=None, end_date=None):
+    def date_filter(record, start_date=None, end_date=None, sampling_date_type=True):
         status = True
 
         if start_date is not None:
-            status = status and record["sampling_end_date"] and start_date <= record["sampling_end_date"]
+            # If sampling_date_type is True, compare to the sampling start date
```
atatmaja, line 167: i think this would be cleaner:

```
record_end = record["sampling_end_date"] if sampling_date_type else record["publication_date"]
```

Then we can basically keep the old logic, just using `record_end` in place of `record["sampling_end_date"]`

app/utils/get_filtered_records.py
```diff
@@ -159,23 +160,38 @@ def should_include(d, k, v):
...
+            if sampling_date_type:
+                status = status and record["sampling_start_date"] and record["sampling_start_date"] >= start_date
```
atatmaja, line 169: also this should be comparing against `sampling_end_date`, not `sampling_start_date`

app/utils/get_filtered_records.py
atatmaja: Also we should make the param name more bool like, maybe `use_sampling_date`

app/utils/helper_funcs.py
```diff
@@ -11,15 +11,21 @@ def validate_request_input_against_schema(input_payload, schema):
     return payload, 200
 
 
-def convert_start_end_dates(data):
-    start_date = data.get('start_date')
+def convert_start_end_dates(data, sampling_date_type=True):
+    # Set the proper date type (sampling or publication date)
```
atatmaja, line 15: Let's also make this request param more bool like

atatmaja (commented):
Good work, a few comments. Also we should block merging this until we can modify front end so that it doesn't break due to request schema changes

app/utils/get_filtered_records.py
simonarocco: are you sure? we want to include the record if it's sampling start date is greater than the minimum start_date that we want, no? e.g. if start_date is Nov 10 2019, and the record sampling_start_date is Dec 10, 2019, we want to check if Dec 10, 2019 > Nov 10 2019

atatmaja: As addressed on slack, my comment holds

app/database_etl/airtable_records_handler/airtable_records_formatter.py
simonarocco: @rahularoradfs what do u think

rahularoradfs: ind_sp_n and ind_se_n should be ints as they are counts - hence why we don't need to divide by 100 to put them on a percentage scale

ind_sp and ind_se are currently percentages expressed in range [0,100] that we are rescaling to [0,1] - so these should be floats

app/namespaces/data_provider/data_provider_schema.py
simonarocco: @brett-stephen

atatmaja (approved):
LGTM but don't merge until we coordinate with brett re front end changes

app/namespaces/data_provider/data_provider_controller.py
```diff
@@ -40,18 +40,25 @@ def post(self):
...
-        result = get_filtered_records(research_fields, filters, columns, start_date=start_date, end_date=end_date,
+        sampling_start_date, sampling_end_date = convert_start_end_dates(data, use_sampling_date=True)
+        publication_start_date, publication_end_date = convert_start_end_dates(data, use_sampling_date=False)
+
+        result = get_filtered_records(research_fields,
+                                      filters,
+                                      columns,
+                                      sampling_start_date=sampling_start_date,
+                                      sampling_end_date=sampling_end_date,
+                                      publication_start_date=publication_start_date,
```
rahularoradfs, line 51: @dtanying - note the addition of publication start date and end date filters here.
You'll want to provide a Jan 1 2020 publication start date and Dec 31 2020 publication end date when you are making requests, to ensure that only data that meets the inclusion criteria for the systematic review is included

PR #155, opened 2021-02-14
OWID Data Ingestion

abeljoseph:
Ingest OWID data (test count, case count, death count, vaccination count, full vaccination count).

https://airtable.com/tbli2lWQHAqBa6ZcI/viw8Ro4zYPYcBGXRw/recgB9XlPVYAm8rj3

Tested applying the migration, running ETL, and executing `data_provider/records` endpt locally. Got response with new fields populated.

No


app/database_etl/owid_ingestion_handler/owid_data_handler.py
```diff
@@ -0,0 +1,84 @@
...
+
+CSV_DIR = os.getenv('CSV_DIR')
+
+vaccination_df = pd.read_csv(CSV_DIR + "vaccinations.csv")[['date', 'iso_code', 'people_vaccinated_per_hundred', 'people_fully_vaccinated_per_hundred']]
+tests_df = pd.read_csv(CSV_DIR + "tests.csv")[['Entity', 'Date', 'ISO code', 'Cumulative total per thousand']]  # per thousand
+cases_df = pd.read_csv(CSV_DIR + "cases.csv")  # per million
+deaths_df = pd.read_csv(CSV_DIR + "deaths.csv")  # per million
+
+
+def _get_alt_name(country_name, df):
```
atatmaja, line 15: nit could add parameter typing here too, was a nice touch in all the other functions

app/database_etl/owid_ingestion_handler/owid_data_handler.py
```diff
@@ -0,0 +1,84 @@
...
+def get_midpoint(start_date: datetime, end_date: datetime):
+    return start_date + (end_date - start_date) / 2
+
+
+def get_total_tests(country_name: str, midpoint_date: datetime):
+    if str(midpoint_date) == 'NaT': return None
+    country_id = get_country_code(country_name)
+    country_total_tests = tests_df.loc[tests_df['ISO code'] == country_id]
+    per_thousand = country_total_tests.loc[country_total_tests['Date'] == midpoint_date.strftime('%Y-%m-%d')][['Entity', 'Cumulative total per thousand']]
+    ret = per_thousand[per_thousand['Entity'].str.contains('tests performed')]['Cumulative total per thousand']
```
atatmaja, line 35: can you explain this logic a bit, not super apparent why we need the str.contains

I'm guessing the tests csv is a bit diff in structure than I expect, if so maybe worthwhile to put a comment above describing it

app/database_etl/location_utils/country_alternative_names.json
```diff
@@ -0,0 +1,30 @@
+{
```
atatmaja, line 1: why do we need 2 alt names? Do the deaths and cases dfs use diff country names???

atatmaja (commented):
Minor comments/questions but overall good work and v clean code!

app/database_etl/location_utils/country_alternative_names.json
abeljoseph: No, but we have different names in Airtable

app/database_etl/location_utils/location_util_functions.py
```diff
@@ -75,6 +75,10 @@ def get_country_code(country_name, iso3=True):
     return code
 
 
+def get_alternative_names(country_code):
+    return COUNTRY_ALTERNATIVE_NAMES[country_code] if country_code in COUNTRY_ALTERNATIVE_NAMES else None
```
simonarocco, line 79: I don't think we need to make a global constant. Just read the file in this function instead of outside

app/database_etl/owid_ingestion_handler/owid_data_handler.py
```diff
@@ -0,0 +1,84 @@
+import pandas as pd
+import datetime
+import os
+
+from app.database_etl.location_utils import get_alternative_names, get_country_code
+
+CSV_DIR = os.getenv('CSV_DIR')
+
+vaccination_df = pd.read_csv(CSV_DIR + "vaccinations.csv")[['date', 'iso_code', 'people_vaccinated_per_hundred', 'people_fully_vaccinated_per_hundred']]
```
simonarocco, line 9: Same thing here - not sure why these need to be global constants. Each of these is only used in one of the functions below. Let's just define these variables in the functions they're used. Makes it harder to understand where these variables are used if they're defined at the top outside the functions

app/database_etl/owid_ingestion_handler/owid_data_handler.py
simonarocco: Agree with this - in general for these functions, would be helpful to have more comments explaining the logic.

app/database_etl/owid_ingestion_handler/owid_data_handler.py
```diff
@@ -0,0 +1,84 @@
...
+def get_total_tests(country_name: str, midpoint_date: datetime):
+    if str(midpoint_date) == 'NaT': return None
+    country_id = get_country_code(country_name)
+    country_total_tests = tests_df.loc[tests_df['ISO code'] == country_id]
+    per_thousand = country_total_tests.loc[country_total_tests['Date'] == midpoint_date.strftime('%Y-%m-%d')][['Entity', 'Cumulative total per thousand']]
+    ret = per_thousand[per_thousand['Entity'].str.contains('tests performed')]['Cumulative total per thousand']
+    return float(ret) / 10 if not ret.empty else None
+
+
+def get_total_cases(country_name: str, midpoint_date: datetime):
```
simonarocco, line 39: e.g. here would be helpful to explain why there is an offset of 9 days for future ref

app/database_etl/owid_ingestion_handler/owid_data_handler.py
```diff
@@ -0,0 +1,84 @@
...
+
+    if country_name not in deaths_df.columns:
+        country_name = _get_alt_name(country_name, deaths_df)
+
+    if country_name is None: return None
+    per_million = deaths_df[['date', country_name]].loc[deaths_df['date'] == target_date][country_name]
+    return float(per_million) / 10000 if not per_million.empty else None
+
+
+def get_vaccinated(country_name: str, midpoint_date: datetime):
```
simonarocco, line 65: can we combine this function and the one below? they are the exact same just for the vaccination field name - can we generalize this function and have a variable that is passed in that specifies the vaccination field name?

simonarocco (changes requested):
good overall! some comments for you to review. did you test that these additional fields are returned when you call the /records endpoint?

app/database_etl/location_utils/location_util_functions.py
abeljoseph: In this case, we need to read the file into memory every time the function is called. It's better design to read it in once, outside the function

app/database_etl/owid_ingestion_handler/owid_data_handler.py
abeljoseph: Again – these functions are called hundreds of times. For each function call, we are going to be using system resources to read the CSV, load it into memory, then have Python garbage collect when the function returns.

I understand that it may difficult to understand where the variables come from. Is there another solution that would allow the code to be clearer, while not eating up resources?

app/database_etl/owid_ingestion_handler/owid_data_handler.py
abeljoseph: Noted – will address.

app/database_etl/owid_ingestion_handler/owid_data_handler.py
abeljoseph: Will address in a comment

app/database_etl/owid_ingestion_handler/owid_data_handler.py
simonarocco: @atatmaja thoughts?

atatmaja: The only other solution I can think of is defining these DFs in `create_dashboard_source_df` and passing them into the owid_handler functions as params. Tbh I think that's more unclear than this global constant approach so I'm okay with this for now @simonarocco @abeljoseph

app/database_etl/owid_ingestion_handler/owid_data_handler.py
```diff
@@ -62,23 +69,13 @@ def get_total_deaths(country_name: str, midpoint_date: datetime):
...
+def get_vaccinated(country_name: str, midpoint_date: datetime, fully_vaccinated=False):
     if str(midpoint_date) == 'NaT': return None
     country_id = get_country_code(country_name)
-    offset = datetime.timedelta(days=-14)
+    offset = OFFSETS['VACCINATIONS']
     target_date = (midpoint_date + offset).strftime('%Y-%m-%d')
 
     country_data = vaccination_df.loc[vaccination_df['iso_code'] == country_id]
-    ret = country_data.loc[country_data['date'] == target_date]['people_fully_vaccinated_per_hundred']
+    ret = country_data.loc[country_data['date'] == target_date]['people_vaccinated_per_hundred'] if not fully_vaccinated else \
```
atatmaja, line 81:
```
col_name = 'people_fully_vaccinated_per_hundred' if fully_vaccinated else 'people_vaccinated_per_hundred'
ret = country_data.loc[country_data['date'] == target_date][col_name]
```

atatmaja: or at least switch the order, if not fully_vaccinated is slightly confusing

atatmaja (approved):
LGTM once tested!

PR #158, opened 2021-02-18
 Updated pagination endpoint

annoyingpr:
Modified `get_pagination_records` to return a dictionary of lists of records (between a start and end page index). Previously, it returned a single page of records. 

[Ticket link](https://airtable.com/tbli2lWQHAqBa6ZcI/viw8Ro4zYPYcBGXRw/receg2IU5tYhWuQzM?blocks=hide)

Made GET and POST requests with the pagination parameters and ensured the behaviour was correct

Nope!

### DIFF
--- a/app/namespaces/data_provider/data_provider_controller.py
+++ b/app/namespaces/data_provider/data_provider_controller.py
@@ -5,7 +5,7 @@ from flask import jsonify, make_response, request
 
 from .data_provider_service import get_record_details, get_country_seroprev_summaries, jitter_pins
 from .data_provider_schema import RecordDetailsSchema, RecordsSchema, StudyCountSchema
-from app.utils import validate_request_input_against_schema, get_filtered_records, get_paginated_records_list, get_paginated_records_dict, convert_start_end_dates
+from app.utils import validate_request_input_against_schema, get_filtered_records, get_paginated_records, convert_start_end_dates
 from app.database_etl.postgres_tables_handler import get_all_filter_options
 
 data_provider_ns = Namespace('data_provider', description='Endpoints for getting database records.')
@@ -37,10 +37,6 @@ class Records(Resource):
             # If there was an error with the input payload, return the error and 422 response
             return make_response(payload, status_code)
 
-        sorting_key = data.get('sorting_key')
-        page_index = data.get('page_index')
-        per_page = data.get('per_page')
-        reverse = data.get('reverse')
         columns = data.get('columns')
         research_fields = data.get('research_fields')
         prioritize_estimates = data.get('prioritize_estimates', True)
@@ -50,10 +46,6 @@ class Records(Resource):
                                       prioritize_estimates=prioritize_estimates)
         if not columns or ("pin_latitude" in columns and "pin_longitude" in columns):
             result = jitter_pins(result)
-
-        # Only paginate if all the pagination parameters have been specified
-        if page_index is not None and per_page is not None and sorting_key is not None and reverse is not None:
-            result = get_paginated_records_list(result, sorting_key, page_index, per_page, reverse)
         return jsonify(result)
 
 @data_provider_ns.route('/records/paginated', methods=['POST'])
@@ -99,7 +91,7 @@ class Records(Resource):
 
         # Only paginate if all the pagination parameters have been specified
         if min_page_index is not None and max_page_index is not None and per_page is not None and sorting_key is not None and reverse is not None:
-            result = get_paginated_records_dict(result, sorting_key, min_page_index, max_page_index, per_page, reverse)
+            result = get_paginated_records(result, sorting_key, min_page_index, max_page_index, per_page, reverse)
         return jsonify(result)
 
 
--- a/app/namespaces/data_provider/data_provider_controller.py
+++ b/app/namespaces/data_provider/data_provider_controller.py
@@ -13,35 +13,9 @@ data_provider_ns = Namespace('data_provider', description='Endpoints for getting
 logging.getLogger(__name__)
 
 
-@data_provider_ns.route('/records', methods=['GET', 'POST'])
+@data_provider_ns.route('/records', methods=['POST'])
 class Records(Resource):
     @data_provider_ns.doc('An endpoint for getting all records from database with or without filters.')
-    def get(self):
-        # Parse pagination request args if they are present
-        sorting_key = request.args.get('sorting_key', None, type=str)
-        min_page_index = request.args.get('min_page_index', None, type=int)
-        max_page_index = request.args.get('max_page_index', None, type=int)
-        per_page = request.args.get('per_page', None, type=int)
-
-        # Type must be string not bool, because bool evaluates to true for any non None value including False and True
-        research_fields = False if str.lower(request.args.get('research_fields', 'false', type=str)) == 'false' else True
-        prioritize_estimates = True if str.lower(request.args.get('prioritize_estimates', 'true', type=str)) == 'true' else False
-        reverse = False if str.lower(request.args.get('reverse', 'false', type=str)) == 'false' else True
-
-        # Log request info
-        logging.info("Endpoint Type: {type}, Endpoint Path: {path}, Arguments: {args}".format(
-            type=request.environ['REQUEST_METHOD'],
-            path=request.environ['PATH_INFO'],
-            args=dict(request.args)))
-
-        result = get_filtered_records(research_fields, filters=None, columns=None, start_date=None, end_date=None,
-                                      prioritize_estimates=prioritize_estimates)
-        result = jitter_pins(result)
-
-        # Only paginate if all the pagination parameters have been specified
-        if min_page_index is not None and max_page_index is not None and per_page is not None and sorting_key is not None and reverse is not None:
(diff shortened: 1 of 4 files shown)


app/utils/get_filtered_records.py
```diff
@@ -246,11 +246,13 @@ def grab_cols(result, columns):
...
 
 
-def get_paginated_records(query_dicts, sorting_key, page_index, per_page, reverse):
+def get_paginated_records(query_dicts, sorting_key, min_page_index, max_page_index, per_page, reverse):
     # Order the records first
     sorted_records = sorted(query_dicts, key=lambda x: (x[sorting_key] is None, x[sorting_key]), reverse=reverse)
 
-    start = page_index * per_page
-    end = page_index * per_page + per_page
+    # Input is non-zero indexing, but we map to zero indexing (e.g. input 1-3 maps to 0-2) but we still return non-zero indexing
```
atatmaja, line 253: Why would we want the input to be non-zero indexing? I think consumers of the endpoint (aka you, prannoy) would prefer zero indexing anyways?

app/utils/get_filtered_records.py
```diff
@@ -246,11 +246,13 @@ def grab_cols(result, columns):
 '''
 
 
-def get_paginated_records(query_dicts, sorting_key, page_index, per_page, reverse):
+def get_paginated_records(query_dicts, sorting_key, min_page_index, max_page_index, per_page, reverse):
```
atatmaja, line 249: I think something we should start moving towards is parameter typing in this repo, this PR would be a nice place to start https://docs.python.org/3/library/typing.html

app/namespaces/data_provider/data_provider_controller.py
```diff
@@ -79,8 +81,8 @@ def post(self):
             result = jitter_pins(result)
 
         # Only paginate if all the pagination parameters have been specified
-        if page_index is not None and per_page is not None and sorting_key is not None and reverse is not None:
-            result = get_paginated_records(result, sorting_key, page_index, per_page, reverse)
+        if min_page_index is not None and max_page_index is not None and per_page is not None and sorting_key is not None and reverse is not None:
```
atatmaja, line 84: Do we need sorting_key and reverse to be specified in order to have to paginate. Maybe we can have default values (e.g. reverse should default to false, sorting_key could maybe default to sampling_end_date???)

app/utils/get_filtered_records.py
simonarocco: We talked about this and we thought it made more sense for the frontend to use non-zero indexing to match the page indices of the table. Like if the frontend wants data for pages 1-3 on the table, it is nicer for them to input indices 1 and 3 instead of 0 and 2.

atatmaja:
@simonarocco @annoyingpr what do you guys think about making pagination its own sep endpoint? I know it does lots of the same stuff as our regular records endpoints but given that the return structure is different, I think it would be nice to make it its own endpoint explicitly

app/utils/get_filtered_records.py
atatmaja: Okay fair enough

atatmaja (commented):
Few comments, good work so far prannoy!

simonarocco:
> @simonarocco @annoyingpr what do you guys think about making pagination its own sep endpoint? I know it does lots of the same stuff as our regular records endpoints but given that the return structure is different, I think it would be nice to make it its own endpoint explicitly

@atatmaja We talked about this a lot. I think the fact that it returns a diff structure is outweighed by the fact that the pagination endpoint would literally be a copy and past of the /records endpoint. We would have to make a /pagination endpoint with both GET and POST methods and it would be the exact same code as the /records. I don't think it's worth it. I know it's kinda of a code smell to have diff return types but I don't think the slight additional functionality of paginating records warrants an entire endpoint of its own.

atatmaja:
> > @simonarocco @annoyingpr what do you guys think about making pagination its own sep endpoint? I know it does lots of the same stuff as our regular records endpoints but given that the return structure is different, I think it would be nice to make it its own endpoint explicitly
> 
> @atatmaja We talked about this a lot. I think the fact that it returns a diff structure is outweighed by the fact that the pagination endpoint would literally be a copy and past of the /records endpoint. We would have to make a /pagination endpoint with both GET and POST methods and it would be the exact same code as the /records. I don't think it's worth it. I know it's kinda of a code smell to have diff return types but I don't think the slight additional functionality of paginating records warrants an entire endpoint of its own.

Couple notes: 

First not sure why we even have a GET and POST given that you can supply filters: {} to the POST in order to replicate the same functionality as the GET.

Second, not too much code would be copy pasted (esp if we only had to replicate the POST endpoint), since most of the actual computation is abstracted into other functions anyways. However, I do still get the sentiment of not wanting to repeat code or maintain things in duplicate.

Third, I feel like the lack of explicitness in what will cause the return structure to change will be a huge source of confusion and potentially errors in the frontend moving down the road. Especially if someone other than Prannoy needs to update any component to use pagination functionality in the future.

At the end of the day I trust your judgement, but these are the things (the third especially) that I feel need to be flagged.

atatmaja:
Need to update `iit-backend/iit-backend/app/utils/__init__.py` to import `get_paginated_records_dict` and `get_paginated_records_list` instead of `get_paginated_records`. That should fix the failing CI/CD check

app/namespaces/data_provider/data_provider_controller.py
```diff
@@ -5,43 +5,61 @@
...
+        start_date, end_date = convert_start_end_dates(data)
+
+        result = get_filtered_records(research_fields, filters, columns, start_date=start_date, end_date=end_date,
                                       prioritize_estimates=prioritize_estimates)
-        result = jitter_pins(result)
+        if not columns or ("pin_latitude" in columns and "pin_longitude" in columns):
+            result = jitter_pins(result)
 
         # Only paginate if all the pagination parameters have been specified
         if page_index is not None and per_page is not None and sorting_key is not None and reverse is not None:
```
atatmaja, line 55: tbh I think we can get rid of this block and the `get_paginated_records_list` function. nobody uses these pagination arguments rn and i think it would be clearer if we have one clearly defined scheme for pagination instead of two

atatmaja (commented):
Looks good, a few comments to address

app/namespaces/data_provider/data_provider_controller.py
annoyingpr: Great point, `/records` should have no pagination functionality - that should only be in `/records/paginated`. Updated!

app/namespaces/data_provider/data_provider_schema.py
```diff
@@ -34,6 +35,14 @@ class RecordsSchema(Schema):
     start_date = fields.String()
     end_date = fields.String()
 
+class PaginatedRecordsSchema(RecordsSchema):
```
simonarocco, line 38: add the param columns to this schema check since it's part of the payload

app/namespaces/data_provider/data_provider_controller.py
```diff
@@ -59,13 +69,14 @@ def post(self):
         filters = data.get('filters')
 
         # Validate input payload
-        payload, status_code = validate_request_input_against_schema(data, RecordsSchema())
+        payload, status_code = validate_request_input_against_schema(data, PaginatedRecordsSchema())
         if status_code != 200:
             # If there was an error with the input payload, return the error and 422 response
             return make_response(payload, status_code)
 
         sorting_key = data.get('sorting_key')
```
simonarocco, line 77: Since sorting_key, per_page, reverse and columns can all be None (based on the schema u defined), this will fail if any of these keys are not in data. We should do data.get(<keyname>, None) (e.g. data.get('sorting_key'None)) so that the variable will default to None if it doesn't exist in the payload

app/namespaces/data_provider/data_provider_controller.py
```diff
@@ -78,9 +89,19 @@ def post(self):
...
+        kwargs = {
+            "records": result,
+            "min_page_index": min_page_index,
+            "max_page_index": max_page_index,
+            "per_page": per_page,
+            "reverse": reverse,
+            "sorting_key": sorting_key
+        }
+
+        kwargs_not_none = { k: v for k, v in kwargs.items() if v is not None }
```
simonarocco, line 101: super nit but can you change the spacing so that it's {k:v for k, v in kwargs.items() if v is not None }

simonarocco (changes requested):
Nice work!! Good comments and typing in the method definitions. just a few changes

app/namespaces/data_provider/data_provider_schema.py
annoyingpr: Just want to make sure, since `PaginatedRecordsSchema` inherits from `RecordsSchema` - shouldn't the `columns` param already be included?

simonarocco: good point! no change needed here then.

app/namespaces/data_provider/data_provider_controller.py
annoyingpr: Great catch, updated this for `sorting_key`, `per_page` and `reverse`. Columns isn't be allowed to take `None`, so I think we should be good for that param.

simonarocco (approved):
lgtm as long as it has been tested

PR #162, opened 2021-02-21
changing path of creds files for google services manager

simonarocco:
- change path of pickle file and credentials.json to explicitly be in tableau_data_connector folder
- need to specify this bc the upload csv to gdrive func is being called outside that folder

### DIFF
--- a/app/database_etl/tableau_data_connector/google_services_manager.py
+++ b/app/database_etl/tableau_data_connector/google_services_manager.py
@@ -19,8 +19,8 @@ class GoogleSheetsManager:
 
     def create_service(self):
         creds = None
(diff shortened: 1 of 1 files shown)


abeljoseph (approved):
LGTM

atatmaja (approved):
LGTM

One thing to consider though, do we know if the token will ever expire? If so we need to put some checks in place to monitor this

simonarocco:
> LGTM
> 
> One thing to consider though, do we know if the token will ever expire? If so we need to put some checks in place to monitor this

Based on the fact that the token for the api scraper for the private sector team hasnt expired yet (and it was generated back in july) i think we're ok.

PR #163, opened 2021-02-23
Update country names to use WHO mapping

abeljoseph:
Update country names using WHO mapping.

https://airtable.com/tbli2lWQHAqBa6ZcI/viw8Ro4zYPYcBGXRw/recotUaIHpSD2vWtp?blocks=hide

Ran the ETL script locally.

No.


app/database_etl/location_utils/country_iso2.json
```diff
@@ -1,59 +1,241 @@
...
+  "Uzbekistan": "UZ",
+  "Vanuatu": "VU",
+  "Venezuela (Bolivarian Republic of)": "VE",
+  "Viet Nam": "VN",
+  "Wallis and Futuna": "WF",
+  "Western Sahara": "EH",
+  "Yemen": "YE",
+  "Zambia": "ZM",
+  "Zimbabwe": "ZW"
 }
```
abeljoseph, line 241: You may notice that there are three less countries compared to `country_iso3.json`. The missing countries are: _Wake Island_, _Midway Islands_, and _Johnston Atoll_.  
  
This is because the codes for these countries have since been deleted (see [ISO 3166-1 alpha-3 Wiki](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3#Deleted_codes)). The ISO2 equivalents are different, and available at this [Wiki page](https://en.wikipedia.org/wiki/ISO_3166-2:UM).  

How should we navigate this? @atatmaja

app/database_etl/location_utils/country_iso2.json
```diff
@@ -1,59 +1,241 @@
...
+  "Comoros": "KM",
+  "Congo": "CG",
+  "Cook Islands": "CK",
+  "Costa Rica": "CR",
+  "Croatia": "HR",
+  "Cuba": "CU",
+  "Cura\u00e7ao": "CW",
+  "Cyprus": "CY",
+  "Czechia": "CZ",
+  "C\u00f4te d'Ivoire": "CI",
```
atatmaja, line 57: Why are accents represented differently in this json file compared to the country_iso3 json file? Have we tested that iso codes for these countries are being looked up properly?

app/database_etl/location_utils/country_iso2.json
atatmaja: Apparently Wake island, Midway Islands, and johnston atoll are now part of the "United States Minor Outlying Islands", which has an iso code of UMI/UM. 

https://www.iso.org/obp/ui/#iso:code:3166:UM

atatmaja (commented):
Thanks for quick turnaround! Some comments.

app/database_etl/location_utils/country_iso2.json
abeljoseph: In Python, they equate to the same thing (the encoded string) so there is no functional difference. However, I can change the country name in this file to match the other one.
image

app/database_etl/location_utils/country_iso2.json
abeljoseph: Yes – the second Wiki page I sent summarizes this.  
My question is: should we keep the old encoding to stay consistent with the ISO3 mapping we have, or should we use the new encoding here?

app/database_etl/location_utils/country_iso2.json
abeljoseph: Country name has been changed.

app/database_etl/location_utils/country_iso2.json
atatmaja: I say we can just map all three to UMI/UM

atatmaja: I think we can push this for now - it's not likely to affect us for a while (don't see us getting a study from any of these places anytime soon). 

Sometime soon, we can follow up with the WHO about this and see if they can revise their list of country name to code mappings. Does this sound like a plan @rahularoradfs

atatmaja (approved):

app/database_etl/location_utils/country_iso2.json
rahularoradfs: Let's not sweat it too much - can follow up with Kathleen at WHO - will forward you a note that you can build off of to follow up if desired!

atatmaja (approved):
LGTM!

PR #164, opened 2021-02-26
add jitter pins to tableau data generator

simonarocco:
- calls jitter pins in the tableau data generator


atatmaja (approved):
LGTM

PR #165, opened 2021-02-26
Add new filter options, add more useful logging to catch addition/removal of filter options

atatmaja:

Add new filter options, add more useful logging to catch addition/removal of filter options. Adds typing to some existing functions.

N/A, addresses the following ETL log:

image

Ran ETL locally, tested get_filter_options endpoint.

No.


simonarocco (approved):
nice! looks good. I like that we're showing the new/old options

PR #167, opened 2021-03-01
fix first_author index out of bounds error

simonarocco:
- in the ETL, we were getting an issue where if there were multiple first authors listed, the index that was computed was longer than the number of first authors, so it threw an index out of bounds error
- this is because for first_author, organizational_author, and source_type, there can be multiple of these three, and we should take the one corresponding to the max publication date
- in some cases, there may not be a first_author corresponding to the max pub date
- example: publication_dates: [2020-03-10, 2020-05,11]. The max pub date corresponds to index 1, so we should get the element of first_author, organizational_author, and source_type where index = 1. If first_author: [person], then first_author[1], is out of bounds. In this case, if the index is out of bounds, get the value with the largest index instead (in this case 0)
- apply this logic to organizational_author and source_type too

- N/A

- ran ETL, made sure no out of bounds error occurred

### DIFF
--- a/app/database_etl/airtable_records_handler/airtable_records_formatter.py
+++ b/app/database_etl/airtable_records_handler/airtable_records_formatter.py
@@ -2,6 +2,7 @@ import numpy as np
 
 from ..location_utils import get_city
 
+
 def get_most_recent_publication_info(row):
     # Get index of most recent pub date if the pub date is not None
     try:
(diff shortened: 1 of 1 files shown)


app/database_etl/airtable_records_handler/airtable_records_formatter.py
```diff
@@ -2,6 +2,7 @@
 
 from ..location_utils import get_city
 
+
 def get_most_recent_publication_info(row):
```
atatmaja, line 6: nit can we add typing to this function

atatmaja (approved):
NVM need to think about the issue a bit more, ignore this approval

atatmaja (approved):
LGTM - just the one suggestion about adding type hints

rahularoradfs:
lgtm :)

rahularoradfs (approved):

PR #168, opened 2021-03-04
remove length limits for research source schema validation

simonarocco:
- Remove length limits for fields in research source table schema validation
- We don't have length limits defined in our models for this table anyways so our validation shouldnt include limits

- Ran ETL and made sure we didnt get the error about not matching validation

- No


atatmaja (approved):
LGTM once you double check that there aren't actually any length limits in the research source tbl

PR #169, opened 2021-03-04
Add flag for exact geographic match, make case/death/vax variable & function names verbose

abeljoseph:
Add a flag for whether a geographic match for case/death/vax data is 1:1, and make variable & function names more verbose.

N/A

Ran the ETL locally, made request to local backend, inspected response.

No.


atatmaja (commented):
Code looks fine, confused about why we need the geo exact match flag now if it'll just always be true. Can't we just implement this when we need?

atatmaja (changes requested):
I'm not sure we've interpreted what @jordanvanwyk meant by `exact_geo_match`.

For ex, say I had a study in Quebec. The canadian vaccination count would not provide an accurate basis of comparison and thus it should not have an `exact_geo_match` right?

app/database_etl/owid_ingestion_handler/owid_data_handler.py
```diff
@@ -90,3 +91,8 @@ def get_vaccinated(country_name: str, midpoint_date: datetime, fully_vaccinated=
     col_name = 'people_fully_vaccinated_per_hundred' if fully_vaccinated else 'people_vaccinated_per_hundred'
     ret = country_data.loc[country_data['date'] == target_date][col_name]
     return float(ret) if not ret.empty else None
+
+
+def get_whether_exact_match(country_name: str, estimate_grade: str):
```
atatmaja98, line 96: let's remove this country_name param if unused for now

atatmaja98, line 96: Suggestion: we could make this function have 2 arguments `record_estimate_grade` and `owid_data_estimate_grade` (probably needs a more appropriate name b/c owid won't always be our source but u get what I mean).

Then the logic would look like `return record_estimate_grade == owid_data_estimate_grade`. Functionally this would act the same way as our current logic. However imo this will make us less likely to forget to update this logic in the future when we do add more granular data. Just a suggestion though, not mandatory.

atatmaja98 (commented):
Looks pretty good, one nit and one suggestion

app/database_etl/owid_ingestion_handler/owid_data_handler.py
abeljoseph: Done.

abeljoseph: Resolving for now – we will need to have a conversation around this mechanism

atatmaja98 (approved):

PR #170, opened 2021-03-07
Add airtable date fields

simonarocco:
- Add last_modified_time and date_created fields from airtable

- Tested ETL works
- Tested uploading to google sheets works

- Apply db migration


app/utils/get_filtered_records.py
```diff
@@ -177,6 +177,13 @@ def date_filter(record, start_date=None, end_date=None):
         if record['sampling_start_date'] is not None:
             record['sampling_start_date'] = record['sampling_start_date'].isoformat()
 
+        # Also apply this formatting to date_created and last_modified_time in research source
+        if research_fields:
```
atatmaja, line 181: minor but I think a safer and more readable way to do this for each of these fields is as follows:

```
isoformat_fields = ['sampling_end_date', 'sampling_start_date', 'date_created', 'last_modified_time']
for field in isoformat_fields:
      if record.get(field, None) is not None:
              record[field] = record[field].isoformat()
```

atatmaja (approved):
minor suggestion, but otherwise LGTM

PR #174, opened 2021-03-11
add additional model options

rahularoradfs:
Austin, when you test these, double check to make sure that there are no warnings on model compilation!


atatmaja:
Noted, will merge and start testing!

PR #180, opened 2021-03-20
Add 'ETL Included' filter to airtable API request

simonarocco:
- Added a new column to airtable called 'ETL Included' that is 1 if Included, Serum pos prevalence, Denominator and Sampling End Date are all not blank, and 0 if any of them are blank
- Added a 'filterByFormula' term to the airtable API request to only pull fields where ETL Included = 1
- Removing code to drop rows where either of these 4 fields is null in the ETL

https://airtable.com/tbli2lWQHAqBa6ZcI/viwp6zUTCqlERLzMj/recMca3RyfhpOQYYO?blocks=hide 

- Checked to make sure that the number of records pulled from airtable when ETL Included = 1 is the same as the number of records without this filter after dropping null fields in ETL

- NO


atatmaja (approved):
Swag as long as we've tested

PR #181, opened 2021-03-20
Test adjustment etl integration

simonarocco:

PR #182, opened 2021-03-21
Automate test evaluation linking

abeljoseph:
Automate the linkage of test evaluations using the ETL.
Criteria found here (page 8): https://docs.google.com/document/d/120ackd8yUTXSbmymL8UniK12dR9ScQONFT_wmG9VQbY

https://airtable.com/tbli2lWQHAqBa6ZcI/viw8Ro4zYPYcBGXRw/recqTLHMSCH27jtJM?blocks=hide

Wrote & ran validation code – checked the results of my algorithm against the existing values (which were added manually based on the criteria).

Results of validation: 100% accuracy

No.


abeljoseph:
## More on validation

Here is the code:
```python
def apply_test_eval_links(df):
    asdf = df[['test_linked_uid', 'ind_eval_link']]
    df['ind_eval_uid_doi'] = df.apply(lambda row: get_eval_UID_and_DOI(row), axis=1)
    asdf['ind_eval_uid_doi'] = df['ind_eval_uid_doi']
    match_count, mismatch_count = 0, 0
    mismatches = pd.DataFrame()
    matches = pd.DataFrame()
    for i, row in asdf.iterrows():
        uid_algo, doi_algo = row['ind_eval_uid_doi']
        if row['test_linked_uid'] is None and row['ind_eval_link'] is None: continue
        try:
            if doi_algo == row['ind_eval_link'] and uid_algo == row['test_linked_uid']:
                matches = matches.append(row)
                match_count += 1
            else:
                mismatches = mismatches.append(row)
                mismatch_count += 1
        except ValueError:
            pass
    df[['test_linked_uid', 'ind_eval_link']] = pd.DataFrame(df['ind_eval_uid_doi'].to_list(), index=df.index)
    df = df.drop(columns=['ind_eval_uid_doi'])
    return df
```   
  
Here are the results:
image

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
```diff
@@ -0,0 +1,100 @@
+import pandas as pd
+
+TEST_EVAL_CSV_LOCATION = 'test_evaluation_handler/test_evaluations.csv'
+
+
+def get_isotype_list(isotype_str):
+    if pd.isna(isotype_str): return []
+    if isotype_str == "Total Ig": return ["IgM", "IgG", "IgA"]
```
atatmaja, line 8: any reason why we're returning a set sometimes and a list other times?

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
```diff
@@ -0,0 +1,100 @@
...
+    isotypes = isotype_str.split('; ')
+    return set(isotypes)
+
+
+test_eval_df = pd.read_csv(TEST_EVAL_CSV_LOCATION)
+test_eval_df['Target'] = test_eval_df.apply(lambda x: get_isotype_list(x['Target']), axis=1)
+manufacturer_mapping = {
```
atatmaja, line 15: worth it to put in a constants file?

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
```diff
@@ -0,0 +1,100 @@
...
+    for i, ev in matches.iterrows():
+        sample_size = ev['SampleSize']
+        if sample_size > largest_sample_size:
+            largest_sample_size, top_match = sample_size, ev.to_frame().T
+
+    return [top_match['unique_id'], top_match['DOI']]
+
+
+def apply_test_eval_links(df):
+    df['ind_eval_uid_doi'] = df.apply(lambda row: get_eval_UID_and_DOI(row), axis=1)
```
atatmaja, line 97: I think we could replace lines 97 to 99 with the following:

```
df['test_linked_uid'], df['ind_eval_link'] = zip(*records_df.apply(lambda row:get_eval_UID_and_DOI(row), axis=1))
```

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
```diff
@@ -0,0 +1,100 @@
...
+
+    # CRITERIA B – sample specimen type
+    specimen_type = row['specimen_type']
+    new_matches = matches[matches['IndexSampleType'] == specimen_type]
+
+    if not new_matches.empty:
+        matches = new_matches
+        if matches.shape[0] == 1: return [matches.iloc[0]['unique_id'], matches.iloc[0]['DOI']]
+
+    # CRITERIA C – prioritized reference specimen type
```
atatmaja, line 61: could you provide a quick comment on what this block is doing at a high lvl

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
```diff
@@ -0,0 +1,100 @@
...
+    if matches.empty: return [None]*2
+
+    if manufacturers[0] == 'Zhejiang Orient Gene (Biotech Co LTD':
+        manufacturers = ['Zhejiang Orient Gene (Biotech Co LTD China)']
+
+    temp = pd.DataFrame()
+    # Get all rows matching any name in manufacturer list
+    for manufacturer in manufacturers:
+        if manufacturer in manufacturer_mapping: manufacturer = manufacturer_mapping[manufacturer]
+        manu = matches[matches['Manufacturer'].str.contains(f"(?i){manufacturer}")]
```
atatmaja, line 41: ?i = caseless?

atatmaja: if so can we comment there

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
```diff
@@ -0,0 +1,100 @@
...
+            if rank == min_rank: top_matches = top_matches.append(ev)  # Same rank, so append
+            else:  # rank < min_rank
+                min_rank, top_matches = rank, ev.to_frame().T
+
+    if not top_matches.empty:
+        matches = top_matches
+        if matches.shape[0] == 1: return [matches.iloc[0]['unique_id'], matches.iloc[0]['DOI']]
+
+    # CRITERIA D – largest sample size
+    largest_sample_size, top_match = 0, pd.DataFrame()
```
atatmaja, line 87: You're just trying to get the row in the matches df with the highest SampleSize here right? You could probably just do this:

```top_match = matches[matches['SampleSize'] ==matches['SampleSize'].max()]```

You'll need to dbl check my syntax here but you get the idea

**Comment by atatmaja** [comment]
> ## More on validation
> Here is the code:
> 
> ```python
> def apply_test_eval_links(df):
>     asdf = df[['test_linked_uid', 'ind_eval_link']]
>     df['ind_eval_uid_doi'] = df.apply(lambda row: get_eval_UID_and_DOI(row), axis=1)
>     asdf['ind_eval_uid_doi'] = df['ind_eval_uid_doi']
>     match_count, mismatch_count = 0, 0
>     mismatches = pd.DataFrame()
>     matches = pd.DataFrame()
>     for i, row in asdf.iterrows():
>         uid_algo, doi_algo = row['ind_eval_uid_doi']
>         if row['test_linked_uid'] is None and row['ind_eval_link'] is None: continue
>         try:
>             if doi_algo == row['ind_eval_link'] and uid_algo == row['test_linked_uid']:
>                 matches = matches.append(row)
>                 match_count += 1
>             else:
>                 mismatches = mismatches.append(row)
>                 mismatch_count += 1
>         except ValueError:
>             pass
>     df[['test_linked_uid', 'ind_eval_link']] = pd.DataFrame(df['ind_eval_uid_doi'].to_list(), index=df.index)
>     df = df.drop(columns=['ind_eval_uid_doi'])
>     return df
> ```
> 
> Here are the results:
> image

Might be worth it to commit this validation code

**Review by atatmaja — COMMENTED** [review]
Nice work Abel! A few comments/suggestions

File: app/database_etl/test_evaluation_handler/test_eval_link_handler.py [hunk 7]
```diff
@@ -0,0 +1,100 @@
...
+test_eval_df['Target'] = test_eval_df.apply(lambda x: get_isotype_list(x['Target']), axis=1)
+manufacturer_mapping = {
+    'Vircell Microbiologists': 'Vircell S.L.',
+    'SNIBE – Shenzhen New Industries Biomedical Engineering Co.': 'Snibe Co., Ltd',
+    'Zhejiang Orient Gene (Biotech Co LTD China)': 'Orient Gene Biotech',
+    'VivaCheck Biotech': 'VivaChek',
+    'Roche Diagnostics': 'Roche'
+}
+
+
```
rahularoradfs, line 23: Overall: do you think it would be worthwhile to abstract the components of this loop[ out a little bit, so folks can see what's going on more clearly? To make clear that you're applying a series of prioritization criteria. The estimate prioritization code provides an example of this. 

Some pseudocode for this could be something like:

```
def get_eval(estimate):
  matches = get_matching_test_manufacturer_name(estimate)
  if not matches: return [None, None]

  for prioritization_function in [prioritize_specimen_type, prioritize_reference_specimen, prioritize_isotype]:
    new_matches = prioritization_function(matches)
    if len(new_matches) == 0: continue
    elif len(new_matches) == 1: return new_matches[0]
    else: matches = new_matches
```

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
```diff
@@ -0,0 +1,100 @@
...
+    temp = pd.DataFrame()
+    # Get all rows matching any name in manufacturer list
+    for manufacturer in manufacturers:
+        if manufacturer in manufacturer_mapping: manufacturer = manufacturer_mapping[manufacturer]
+        manu = matches[matches['Manufacturer'].str.contains(f"(?i){manufacturer}")]
+        temp = temp.append(manu)
+
+    matches = temp
+    # Check that targets are exact match
+    for i, ev in matches.iterrows():
```
rahularoradfs, line 46: `iterrows` and `drop` workflow never usually ideal in pandas. You could do this one better with a vectorized approach. Something like:

matches = matches[matches.apply(lambda match: set(match['Target']) == target_isotypes]

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
```diff
@@ -0,0 +1,100 @@
...
+
+    if not new_matches.empty:
+        matches = new_matches
+        if matches.shape[0] == 1: return [matches.iloc[0]['unique_id'], matches.iloc[0]['DOI']]
+
+    # CRITERIA C – prioritized reference specimen type
+    rankings = {'Lower respiratory specimen': 1, 'Sputum': 3, 'Bronchial aspirate': 3,
+                'Respiratory specimen': 4, 'Upper respiratory specimen': 4, 'Nasal mid-turbinate swabs': 5,
+                'Nasopharyngeal swabs': 7, 'Swab': 7, 'Serum': 8, 'Pharyngeal swabs': 10, 'Plasma': 11, 'Oropharyngeal swabs': 12,
+                'Whole blood': 13, 'Urine': 14, 'Not available': 20, 'not reported': 20, 'Not reported': 20}
```
rahularoradfs, line 65: any reason you're not using `pd.na` for the ones coded 20 here?

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
```diff
@@ -0,0 +1,100 @@
...
+
+    min_rank, top_matches = 100, pd.DataFrame()
+    for i, ev in matches.iterrows():
```
rahularoradfs, line 68: Similarly, you can do this better by removing the iterrows. 

matches.apply(lambda match: rankings.get(match['ReferenceSampleType'], pd.na).idxmax() gets the index of the top element, and you can just index by it

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
```diff
@@ -0,0 +1,100 @@
...
+        if pd.isna(ev['ReferenceSampleType']):
+            if top_matches.empty or min_rank == 20:
+                min_rank = 20
+                top_matches = top_matches.append(ev)
+                continue
+        reference_specimen_types = ev['ReferenceSampleType'].split('; ')
+        for ref in reference_specimen_types:
```
rahularoradfs, line 75: code above takes care of finding a max

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
rahularoradfs: idxmax could be better suited for the inside of the square brackets here- see https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.idxmax.html

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
```diff
@@ -0,0 +1,100 @@
...
+        if matches.shape[0] == 1: return [matches.iloc[0]['unique_id'], matches.iloc[0]['DOI']]
+
+    # CRITERIA D – largest sample size
+    largest_sample_size, top_match = 0, pd.DataFrame()
+    for i, ev in matches.iterrows():
+        sample_size = ev['SampleSize']
+        if sample_size > largest_sample_size:
+            largest_sample_size, top_match = sample_size, ev.to_frame().T
+
+    return [top_match['unique_id'], top_match['DOI']]
```
rahularoradfs, line 93: What if two matches have the same sample size? Should add logic to deal with this

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def apply_test_eval_links(df):
+    df['ind_eval_uid_doi'] = df.apply(lambda row: get_eval_UID_and_DOI(row), axis=1)
+    df[['test_linked_uid', 'ind_eval_link']] = pd.DataFrame(df['ind_eval_uid_doi'].to_list(), index=df.index)
```
rahularoradfs, line 98: @abeljoseph -- you link the test in, but you don't actually pull the corresponding values through here. 

The point of the test link is it lets us pull the independent evaluation values for sensitivity and specificity (in particular, `ind_se`, `ind_se_n`, `ind_sp`, `ind_sp_n`) from the FindDx CSV through to each estimate. We've been doing that with a lookup field in Airtable, but now we should just be able to do a direct lookup from the CSV, right? Once we have done this, we should add these to the dataframe and database, and overwrite the values for those fields that are coming from the Airtable. 

@atatmaja, could you confirm that no other fields are needed for test adjustment here?

rahularoradfs (changes requested):
Great start. @abeljoseph left a fair bit of feedback that will help ensure that this meets functionality needs and is as parsimonious as possible. 

@atatmaja code review might be worth a quick skim from you too to make sure we're on same page with spec and Pandas code quality points. Will leave subsequent review to you!

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
```diff
@@ -0,0 +1,100 @@
+import pandas as pd
+
+TEST_EVAL_CSV_LOCATION = 'test_evaluation_handler/test_evaluations.csv'
+
+
+def get_isotype_list(isotype_str):
+    if pd.isna(isotype_str): return []
+    if isotype_str == "Total Ig": return ["IgM", "IgG", "IgA"]
+    isotypes = isotype_str.split('; ')
+    return set(isotypes)
```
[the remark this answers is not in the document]
atatmaja: yeah we could either do that or just always return a set object

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
```diff
@@ -0,0 +1,100 @@
...
+    'SNIBE – Shenzhen New Industries Biomedical Engineering Co.': 'Snibe Co., Ltd',
+    'Zhejiang Orient Gene (Biotech Co LTD China)': 'Orient Gene Biotech',
+    'VivaCheck Biotech': 'VivaChek',
+    'Roche Diagnostics': 'Roche'
+}
+
+
+def get_eval_UID_and_DOI(row):
+    if not row['test_manufacturer'] or not row['isotypes']: return [None]*2
+    if row['ind_eval_link'] or row['test_linked_uid']: return [row['test_linked_uid'], row['ind_eval_link']]
```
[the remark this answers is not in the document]
atatmaja: If we're sure that it won't be edited, I think we could remove the existing test link table...

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
atatmaja: Okay with idxmax if the use case is that we for sure want to return one row. But the approach I provided can also be used if we want to return all rows that feature a max value for some column, so I think it's a bit more extensible there. 

Either or is a good option for me

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
atatmaja: I think that is the last prioritization criterion we've defined? Did we ever reach this case and end up with multiple results during the systematic review? If so how did we deal with them then? @rahularoradfs

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
atatmaja: I think we also have a column called 'test_validation', not sure if it's populated by test linking @rahularoradfs 

Some context on how it's being used in the test adjustment script:

https://github.com/serotracker/iit-backend/pull/181/files#diff-22acfea80e41338a5fa26a4264e9cce8c2d39121f24e1605d699f56f8c643a43R179

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
abeljoseph: Changed to always return list

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
```diff
@@ -0,0 +1,100 @@
...
+
+def get_isotype_list(isotype_str):
+    if pd.isna(isotype_str): return []
+    if isotype_str == "Total Ig": return ["IgM", "IgG", "IgA"]
+    isotypes = isotype_str.split('; ')
+    return set(isotypes)
+
+
+test_eval_df = pd.read_csv(TEST_EVAL_CSV_LOCATION)
+test_eval_df['Target'] = test_eval_df.apply(lambda x: get_isotype_list(x['Target']), axis=1)
```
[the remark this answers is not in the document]
abeljoseph: Yup 'Target' is right from Airtable. Will add a comment

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
```diff
@@ -0,0 +1,100 @@
...
+manufacturer_mapping = {
+    'Vircell Microbiologists': 'Vircell S.L.',
+    'SNIBE – Shenzhen New Industries Biomedical Engineering Co.': 'Snibe Co., Ltd',
+    'Zhejiang Orient Gene (Biotech Co LTD China)': 'Orient Gene Biotech',
+    'VivaCheck Biotech': 'VivaChek',
+    'Roche Diagnostics': 'Roche'
+}
+
+
+def get_eval_UID_and_DOI(row):
```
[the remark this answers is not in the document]
abeljoseph: Done.

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
rahularoradfs: don't think we did. honestly at that point flip a coin, pick in alphabetical order, lower unique_ID. it's an edge case and shouldn't ever really happen, we should just make sure it's handled in code

app/database_etl/test_evaluation_handler/test_eval_link_handler.py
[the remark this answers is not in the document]
abeljoseph: Returning a list now

PR #184, opened 2021-03-22
WIP: ETL summary report basic features

atatmaja:

Sends a summary report detailing each ETL run to slack. For now we've included the following basic functionality:
- Date & time ETL ran
- ETL runtime
- Record count in each table before and after 
- Record count from Airtable

https://airtable.com/tbli2lWQHAqBa6ZcI/viw8iBsG4yooIYvRI/recaU6vCNWIszwnCI

- Tested locally, made sure that correct summary reports were sent to slack
- Specifically, tested that the summary report would be generated even if an error was raised midway through the ETL, causing it to abort

N/A

### DIFF
--- a/app/database_etl/etl_main.py
+++ b/app/database_etl/etl_main.py
@@ -36,7 +36,7 @@ def main():
 
         # Get all records with airtable API request and load into dataframe
         json = get_all_records()
-        etl_report.get_num_airtable_records(len(json))
+        etl_report.set_num_airtable_records(len(json))
         data = pd.DataFrame(json)
 
         # Clean raw airtable records to standardize data formats
@@ -97,7 +97,7 @@ def main():
         # If all tables were successfully loaded, drop old entries
         if load_status:
             drop_table_entries(current_time=CURR_TIME, drop_old=True)
-            etl_report.get_table_counts_after()
+            etl_report.set_table_counts_after()
         # Otherwise drop entries from current ETL run
         else:
             drop_table_entries(current_time=CURR_TIME, drop_old=False)
--- a/app/database_etl/etl_main.py
+++ b/app/database_etl/etl_main.py
@@ -12,6 +12,7 @@ from app.database_etl.postgres_tables_handler import create_dashboard_source_df,
 from app.database_etl.airtable_records_handler import get_all_records, apply_study_max_estimate_grade,\
     apply_min_risk_of_bias, standardize_airtable_data
 from app.database_etl.tableau_data_connector import upload_analyze_csv
+from app.database_etl.summary_report_generator import SummaryReport
 
 
 load_dotenv()
@@ -30,8 +31,12 @@ def main():
         password=os.getenv('DATABASE_PASSWORD'),
         host_address=os.getenv('DATABASE_HOST_ADDRESS')))
 
+    # Instantiate summary report object
+    etl_report = SummaryReport()
+
     # Get all records with airtable API request and load into dataframe
     json = get_all_records()
+    etl_report.get_num_airtable_records(len(json))
     data = pd.DataFrame(json)
 
     # Clean raw airtable records to standardize data formats
@@ -92,6 +97,7 @@ def main():
     # If all tables were successfully loaded, drop old entries
     if load_status:
         drop_table_entries(current_time=CURR_TIME, drop_old=True)
+        etl_report.get_table_counts_after()
     # Otherwise drop entries from current ETL run
     else:
         drop_table_entries(current_time=CURR_TIME, drop_old=False)
@@ -101,6 +107,10 @@ def main():
 
     # Upload tableau csv to google sheets
     upload_analyze_csv()
+
+    # Send ETL summary report
+    etl_report.get_elapsed_time()
+    etl_report.send_summary_report()
     return
 
 
--- a/app/database_etl/etl_main.py
+++ b/app/database_etl/etl_main.py
@@ -25,93 +25,89 @@ CURR_TIME = datetime.now()
 
 
 def main():
-    # Create engine to connect to whiteclaw database
-    engine = create_engine('postgresql://{username}:{password}@{host_address}/whiteclaw'.format(
-        username=os.getenv('DATABASE_USERNAME'),
-        password=os.getenv('DATABASE_PASSWORD'),
-        host_address=os.getenv('DATABASE_HOST_ADDRESS')))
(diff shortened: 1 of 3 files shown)


app/database_etl/summary_report_generator/summary_report.py
```diff
@@ -0,0 +1,64 @@
...
+
+class SummaryReport:
+    def __init__(self):
+        self.start_time = time()
+        self.get_table_counts_before()
+
+    def get_elapsed_time(self):
+        return time() - self.start_time
+
+    def get_num_airtable_records(self, num_records: int):
```
simonarocco, line 18: shouldn't this be set_num_airtable_records?

app/database_etl/summary_report_generator/summary_report.py
```diff
@@ -0,0 +1,64 @@
...
+                table_counts[table.__tablename__] = session.query(func.count(getattr(table, pk))).scalar()
+        return table_counts
+
+    def get_table_counts_before(self):
+        self.table_counts_before = self.get_table_counts()
+
+    def get_table_counts_after(self):
+        self.table_counts_after = self.get_table_counts()
+
+    def get_human_readable_time(self, timestamp: float):
```
simonarocco, line 39: i feel like we dont rly need this func cause it's just used once and we can put the logic where it's used

app/database_etl/summary_report_generator/summary_report.py
```diff
@@ -0,0 +1,64 @@
...
+            body += f"Status: SUCCESS"
+        else:
+            body += f"Status: FAIL"
+        send_slack_message(body, channel="#dev-logging-etl")
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exception_type, exception_value, traceback):
+        self.get_elapsed_time()
```
simonarocco, line 60: why is this needed? i dont think this function call is needed for getting the elapsed time?

app/database_etl/summary_report_generator/summary_report.py
```diff
@@ -0,0 +1,64 @@
...
+    def get_table_counts(self):
+        table_counts = {}
+        with db_session() as session:
+            for table in [DashboardSource, ResearchSource, Country, State, City,
+                          TestManufacturer, AntibodyTarget, CityBridge, StateBridge,
+                          TestManufacturerBridge, AntibodyTargetBridge]:
+                # Inspect primary key of table
+                pk = inspect(table).primary_key[0].key
+                # Use primary key to get row count instead of using the .count() method for perf reasons
+                table_counts[table.__tablename__] = session.query(func.count(getattr(table, pk))).scalar()
```
simonarocco, line 30: does this work instead? `select([func.count()]).select_from(table)`? reference: https://stackoverflow.com/questions/12941416/how-to-count-rows-with-select-count-with-sqlalchemy

app/database_etl/summary_report_generator/summary_report.py
```diff
@@ -0,0 +1,64 @@
...
+        return table_counts
+
+    def get_table_counts_before(self):
```
simonarocco, line 33: these two functions should be set_

app/database_etl/summary_report_generator/summary_report.py
```diff
@@ -0,0 +1,64 @@
...
+        self.table_counts_after = self.get_table_counts()
+
+    def get_human_readable_time(self, timestamp: float):
+        dt = datetime.datetime.fromtimestamp(timestamp)
+        return dt.strftime('%Y-%m-%d %H:%M:%S')
+
+    def send_summary_report(self):
+        body = f"Summary report for ETL run at {self.get_human_readable_time(self.start_time)}:\n\n"
+        body += f"Total runtime: {self.get_elapsed_time()} seconds\n"
+        if hasattr(self, 'num_airtable_records'):
```
simonarocco, line 46: why would this attribute not get set? it gets set in the ETL no matter what (i.e. not based on a certain condition)

app/database_etl/etl_main.py
```diff
@@ -24,84 +25,89 @@
...
-
-    # Make sure that filter options are still valid
-    check_filter_options(dashboard_source)
-
-    # Upload tableau csv to google sheets
-    upload_analyze_csv()
-    return
+    # etl_report.__exit__() will run whether or not this block is successfully completed
+    # see https://www.geeksforgeeks.org/with-statement-in-python/
+    with SummaryReport() as etl_report:
```
simonarocco, line 30: is there a benefit to making this a generator? is it that the summary report will automatically be sent upon cleanup so we don't have to call the send summary report function?

simonarocco (commented):
looks good!! thanks for this. some overall questions and comments.

app/database_etl/summary_report_generator/summary_report.py
atatmaja: yes agree

app/database_etl/summary_report_generator/summary_report.py
atatmaja: down for abstraction, jk ya sure

app/database_etl/summary_report_generator/summary_report.py
atatmaja: what if the etl fails before we are able to get a number of airtable records (e.g. if the request to airtable fails)

app/database_etl/etl_main.py
atatmaja: yes, basically want to make it so that the send summary report function will always run even if we run into an error midway thru the code

app/database_etl/summary_report_generator/summary_report.py
atatmaja: agree

app/database_etl/summary_report_generator/summary_report.py
atatmaja: it does but i am more down to use ORM layer, more consistent with the rest of our sqlalchemy usage

app/database_etl/summary_report_generator/summary_report.py
atatmaja: u r correct

app/database_etl/summary_report_generator/summary_report.py
atatmaja: actually I am going to leave this as a function here, it makes the code in send_summary_report look cleaner

simonarocco (approved):
gg

PR #186, opened 2021-04-05
Add subgroup filter options

rahularoradfs:

Adding `subgroup_var`, `subgroup_cat`, and `genpop` as valid filter options for the `data_provider/records` endpoint for research purposes 

None

Little testing beside code checks passing - not clear what would be necessary to actually test this functionality! 

Shouldn't be required -- modifications to existing code

### DIFF
--- a/app/database_etl/postgres_tables_handler/postgres_utils.py
+++ b/app/database_etl/postgres_tables_handler/postgres_utils.py
@@ -78,7 +78,7 @@ def get_filter_static_options() -> Dict[str, List[str]]:
         "genpop": [
             'Study examining general population seroprevalence',
             'Study examining special population seroprevalence',
-        ]
+        ],
         "subgroup_var": [
            'Geographical area', 'Health care worker exposure', 'Population',
            'Primary Estimate', 'Comorbidities', 'Sex/Gender',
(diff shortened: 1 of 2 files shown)


app/database_etl/postgres_tables_handler/postgres_utils.py
```diff
@@ -74,7 +74,29 @@ def get_filter_static_options() -> Dict[str, List[str]]:
             "ELISA",
             "LFIA",
             "Other"
-        ]
+        ],
+        "genpop": [
```
atatmaja, line 78: Just want to note that the only reason some of these filter options are hardcoded is because they need to be shown in a certain order on serotracker.com.

That being said, do we ever want to show these filter options on the site / do they ever need to be ordered in a specific way? If not we could just use a select distinct query in `get_all_filter_options` below to get the options for these filters. Example code: 

```
query = session.query(distinct(ResearchSource.genpop))
results = [q[0] for q in query if q[0] is not None]
# sort countries in alpha order
options["genpop"] = sorted(results)
```

atatmaja (commented):
Code looks good, just one comment

app/database_etl/postgres_tables_handler/postgres_utils.py
rahularoradfs: This is a great idea. No reason for us to hardcode this. Will modify to generate filter options dynamically for these variables.

atatmaja (approved):
LGTM as long as we've tested filter options endpoint

PR #189, opened 2021-04-08
Remove superceded column

simonarocco:
- The formula in airtable will now exclude all fields where superceded is true
- This means we no longer need the superceded col in our db cause they will all have the same value
- I also modified the logic for the test adj model integration to skip over test adjustment if there are no new records


atatmaja:
LGTM, let's avoid merging this until we've figured out the problems with our ETL on prod

atatmaja (approved):
LGTM

PR #190, opened 2021-04-09
Add include in srma col and filter

atatmaja:

Adds the "included in SRMA (mar 2021)" column to our ResearchSource table and lets consumers of the /records endpoint filter by it.

https://airtable.com/tbli2lWQHAqBa6ZcI/viwWvxb27XGuziw4c/recrpePqi1G7iZj9z?blocks=bipIkdtYHmboUfJut

- Checked local postgres after ETL run, ensured that number of records with "include_in_srma" = true matched airtable
- Hit /records endpoint locally with "include_in_srma" = true, ensured that the right number of records were returned

Depends on the superceded PR


app/utils/airtable_fields_config.py
```diff
@@ -159,7 +159,6 @@
     'JBI 7': 'JBI_7',
     'JBI 8': 'JBI_8',
     'JBI 9': 'JBI_9',
-    'Superceded?': 'SUPERCEDED',
```
rahularoradfs, line 162: Can get rid of this side of things now - we don't need the old variable names as we don't use them anywhere on research, now fully deprecated

app/utils/get_filtered_records.py
```diff
@@ -128,7 +128,7 @@ def process_record(record_list):
 def get_filtered_records(research_fields=False, filters=None, columns=None,
                          sampling_start_date=None, sampling_end_date=None,
                          publication_start_date=None, publication_end_date=None, prioritize_estimates=True,
-                         prioritize_estimates_mode=None):
+                         prioritize_estimates_mode=None, include_in_srma=False):
```
rahularoradfs, line 131: To me, the False default is a bit unintuitive here - a None default would seem to make more sense. At your discretion though since you've been much more involved in API design, fine with either

rahularoradfs (approved):
lgtm!!

app/utils/get_filtered_records.py
atatmaja: I think True/False is more intuitive than True/None so I am leaving it

app/utils/airtable_fields_config.py
atatmaja: i think simona got rid of this in a sep PR already. I just build my branch off of it so the changes were reflected in the diff when her changes hadn't been merged in yet

PR #191, opened 2021-04-11
Use SQLAlchemy instead of python to aggregate records

atatmaja:

Use sqlalchemy instead of python to perform record aggregation (after joining) to increase code simplicity

N/A

-Hit endpoint locally with certain filter combinations and ensured that the same number of records were retrieved (comparing against making sql queries on my local database)
- Pointed local serotracker.com web client to local server and made sure that all functionality worked as expected

- N/A

### DIFF
--- a/app/utils/get_filtered_records.py
+++ b/app/utils/get_filtered_records.py
@@ -1,13 +1,47 @@
-from itertools import groupby
-from functools import reduce
 from app.serotracker_sqlalchemy import db_session, DashboardSource, ResearchSource, \
     db_model_config, Country, State, City, dashboard_source_cols, research_source_cols
+from sqlalchemy.dialects.postgresql import array
+from sqlalchemy import func, cast, case, and_, Float, String, ARRAY
 import pandas as pd
 import numpy as np
 from app.utils.estimate_prioritization import get_prioritized_estimates
 from statistics import mean
 from typing import List, Dict, Any
 
+def _get_isotype_col_expression(label="isotypes"):
+    expression = case(
+                [
+                    (and_(DashboardSource.isotype_igg == 'true',
+                          DashboardSource.isotype_igm == 'true',
+                          DashboardSource.isotype_iga == 'true'), 'IgG, IgM, IgA'),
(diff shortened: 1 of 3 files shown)


app/utils/get_filtered_records.py
```diff
@@ -206,24 +210,19 @@ def date_filter(record, start_date=None, end_date=None, use_sampling_date=True):
...
-                if 'pin_latitude' not in record:
-                    record['pin_latitude'] = None
-                    record['pin_longitude'] = None
-                    record['pin_region_type'] = ''
-                break
-            # If we find multiple coordinates for a given geographic level
+        record['pin_latitude'] = record['country_latitude']
+        record['pin_longitude'] = record['country_longitude']
+        record['pin_region_type'] = 'country' if record['country_latitude'] is not None else ''
+        for region_type in ['state', 'city']:
```
simonarocco, line 216: no more country region type?

simonarocco (commented):
good overall. close to done. minor nit

app/utils/get_filtered_records.py
```diff
@@ -1,13 +1,47 @@
...
     db_model_config, Country, State, City, dashboard_source_cols, research_source_cols
+from sqlalchemy.dialects.postgresql import array
+from sqlalchemy import func, cast, case, and_, Float, String, ARRAY
 import pandas as pd
 import numpy as np
 from .estimate_prioritization import get_prioritized_estimates
 from statistics import mean
 from typing import List, Dict, Any
 
+def _get_isotype_col_expression(label="isotypes"):
```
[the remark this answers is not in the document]
atatmaja: it's being used in the data provider service and in the `get_all_records` in this function

app/utils/get_filtered_records.py
atatmaja: Lines 213 - 215 set the pin_latitude/longitude to the country latitude/longitude to begin with. Just thought it was cleaner to write it this way, esp because state and city lat/lngs are lists while country lat/lngs are single values.

simonarocco (approved):
guccii

PR #193, opened 2021-04-17
Fix EP Internal Server Errors

atatmaja:

- Bug: internal server errors when calling /records endpoint
- Bugfix: Drop null entries in pooling function so that string concatenation won't fail

N/A

Tested locally (using Postman) with various filter combinations (no filters, country = canada, population group = residual sera, and combinations)

N/A


rahularoradfs (approved):
`dropna()`

atatmaja:
> `dropna()`

all hail dropna

PR #196, opened 2021-04-20
Round all floats to equal precision to ensure consistent comparisons during testadj diffing

atatmaja:

Round all floats to equal precision to ensure consistent comparisons during testadj diffing

N/A

Ran ETL locally

Re-enable CI/CD

### DIFF
--- a/app/database_etl/airtable_records_handler/airtable_records_formatter.py
+++ b/app/database_etl/airtable_records_handler/airtable_records_formatter.py
@@ -130,9 +130,7 @@ def add_test_adjustments(df: pd.DataFrame) -> pd.DataFrame:
     diff[float_cols] = diff[float_cols].astype(float)
 
     # Round float columns to a consistent number of decimal places to ensure consistent float comparisons
-    # Defines number of decimal places to round each column to
-    decimals = pd.Series([5]*len(float_cols), index=float_cols)
-    diff = diff.round(decimals)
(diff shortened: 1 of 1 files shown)


app/database_etl/airtable_records_handler/airtable_records_formatter.py
```diff
@@ -129,6 +129,11 @@ def add_test_adjustments(df: pd.DataFrame) -> pd.DataFrame:
                   'denominator_value', 'serum_pos_prevalence']
     diff[float_cols] = diff[float_cols].astype(float)
 
+    # Round float columns to a consistent number of decimal places to ensure consistent float comparisons
+    # Defines number of decimal places to round each column to
+    decimals = pd.Series([5]*len(float_cols), index=float_cols)
```
rahularoradfs, line 134: gotta admit I had to look up what this was doing! makes sense on lookup.

an alternative way to factor this would be something like
diff[float_cols] = diff[float_cols].round(5)

rahularoradfs (approved):

app/database_etl/airtable_records_handler/airtable_records_formatter.py
atatmaja: think this is cleaner, will refactor

simonarocco (approved):
lgtm. i tested with postgres v13 and also works with that

PR #197, opened 2021-04-20
Adding handling of edge case where there are no old records

simonarocco:
- if there are no old records in db, should just end the test adj function early and return all new adjusted records


rahularoradfs:
Are there any cases where `old_airtable_test_adj_records` might not be a `pd.DataFrame`? @atatmaja and I ran into an issue with that very thing over the course of the past week, so flagging it here. If not, fine by me!

atatmaja:
> Are there any cases where `old_airtable_test_adj_records` might not be a `pd.DataFrame`? @atatmaja and I ran into an issue with that very thing over the course of the past week, so flagging it here. If not, fine by me!

Old airtable records is defined by applying a filter to a DF so I think it should always be one. I think this PR lgtm!

atatmaja (approved):

PR #203, opened 2021-04-22
fix get record details bug

atatmaja:

- Bug: Get record details was returning an array of records instead of a single one. This was causing some record detail modals on the explore tab to appear buggy.
- Solution: Remove pythonic isotype string to list conversion and instead, have sqlalchemy query return an array right away

N/A

- Tested endpoints on postman
- Ran local app and performed regression testing

- N/A

### DIFF
--- a/app/utils/get_filtered_records.py
+++ b/app/utils/get_filtered_records.py
@@ -18,25 +18,25 @@ def _get_isotype_col_expression(label:str = "isotypes"):
                 [
                     (and_(DashboardSource.isotype_igg == 'true',
                           DashboardSource.isotype_igm == 'true',
-                          DashboardSource.isotype_iga == 'true'), array(['IgG, IgM, IgA'])),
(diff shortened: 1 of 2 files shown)


app/utils/get_filtered_records.py
```diff
@@ -18,25 +18,25 @@ def _get_isotype_col_expression(label:str = "isotypes"):
                 [
                     (and_(DashboardSource.isotype_igg == 'true',
                           DashboardSource.isotype_igm == 'true',
-                          DashboardSource.isotype_iga == 'true'), array(['IgG, IgM, IgA'])),
```
rahularoradfs, line 21: I wonder whether there's a better way to do this... 

if DashboardSource.isotype_igx: array.append('igx')
the 8x arrays doesn't seem like the best of design patterns to me!!

this is less of a comment on this PR and more of a general comment

rahularoradfs (approved):
overall lgtm, i don't understand the full backend arch as well as @simonarocco so her review might be helpful

PR #205, opened 2021-04-25
Clarify get filtered records

rahularoradfs:

So the filtering logic in `get_filtered_records` was not particularly readable. I've had to come back to it a few times to understand how it works properly. This time, I figured I'd refactor it and change variable naming to make it easier to understand.

To review, it's probably easier to look at the old and new version of the files rather than at a line by line diff. 

None.

- Loaded local backend
- Tested analysis code on local backend to check whether same results were emitted
- Tested frontend on local backend, testing various combinations of filters and confirming that they work as expected

No.

### DIFF
--- a/app/utils/get_filtered_records.py
+++ b/app/utils/get_filtered_records.py
@@ -144,33 +144,25 @@ def get_filtered_records(research_fields=False, filters=None, columns=None,
     if query_dicts is None or len(query_dicts) == 0:
         return []
 
-    result = []
-
     # Return all records if no filters are passed in
+    result = query_dicts
+
     if filters:
-        def should_include(d, k, v):
-            if len(v) == 0:
+        def should_be_included(record, field, permitted_values):
+            if len(permitted_values) == 0:
+                # all values for field permitted
                 return True
-            elif isinstance(d[k], str) and d[k] in v:
+            elif (isinstance(record[field], str) and 
+                  record[field] in permitted_values):
                 return True
-            elif isinstance(d[k], list) and set(d[k]).intersection(set(v)):
+            elif (isinstance(record[field], list) and 
+                  set(record[field]).intersection(set(permitted_values))):
                 return True
             return False
 
-        applied_filter = False
-        for k, v in filters.items():
-            # Add records passing the first filter
-            if not applied_filter:
-                for d in query_dicts:
-                    if should_include(d, k, v):
-                        result.append(d)
-                applied_filter = True
-                continue
-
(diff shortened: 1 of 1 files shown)


app/utils/get_filtered_records.py
```diff
@@ -144,33 +144,22 @@ def get_filtered_records(research_fields=False, filters=None, columns=None,
     if query_dicts is None or len(query_dicts) == 0:
```
atatmaja, line 144: I think we can be sure query_dicts will never be None, so we can remove that from this conditional

app/utils/get_filtered_records.py
```diff
@@ -144,33 +144,22 @@ def get_filtered_records(research_fields=False, filters=None, columns=None,
...
 
-    result = []
-
     # Return all records if no filters are passed in
+    result = query_dicts
+
     if filters:
-        def should_include(d, k, v):
-            if len(v) == 0:
+        def should_be_included(record, field, permitted_values):
```
atatmaja, line 151: don't see a clear reason as to why we defined these functions as closures. let's move this and the `date_filter` function outside of `get_filtered_records `

atatmaja: And we can add typing while we are at it

app/utils/get_filtered_records.py
```diff
@@ -144,33 +144,22 @@ def get_filtered_records(research_fields=False, filters=None, columns=None,
     if query_dicts is None or len(query_dicts) == 0:
         return []
 
-    result = []
-
     # Return all records if no filters are passed in
+    result = query_dicts
+
     if filters:
```
atatmaja, line 150: @simonarocco do you think it would be cleaner to define all this filtering logic in SQL instead of python? 

I think we should definitely move date filtering to SQLAlchemy cause it's pretty simple, but not too sure about this.

app/utils/get_filtered_records.py
simonarocco: agreed, move outside and add typing

app/utils/get_filtered_records.py
simonarocco: I'm not a fan of this python filtering logic at all, it's hard to read and seems over engineered when there is a better way to do filtering. I agree we should do filtering logic in SQL if we can. I can help @rahularoradfs turn this into sqlalchemy

simonarocco (changes requested):
change python filtering to sqlalchemy

atatmaja:
I think this is stale so I am closing this

PR #206, opened 2021-04-26
augment country_alternative_names.json

rahularoradfs:

Case and test data for some countries was not being pulled because these countries had different names in our data and in the Hopkins data. Augmented the mapping in our data to fix this. 

None. However, I have added a ticket to ensure that countries with missing data are identified automatically via Slack going forward. Ticket link: https://airtable.com/tbli2lWQHAqBa6ZcI/viw8Ro4zYPYcBGXRw/recPEw7IDgiOWalqQ?blocks=hide

Ran ETL locally. 

No.

### DIFF
--- a/app/database_etl/location_utils/country_alternative_names.json
+++ b/app/database_etl/location_utils/country_alternative_names.json
@@ -26,5 +26,22 @@
     "EST": [
         "Republic of Estonia",
         "Estonia"
+    ],
+    "GBR": [
+        "The United Kingdom",
+        "United Kingdom"
+    ],
+    "IRN": [
+        "Iran (Islamic Republic of)",
+        "Iran"
+    ], 
+    "JEY": [
+        "Jersey"
+    ],
+    "GUF": [
+        "French Guiana"
+    ],
+    "FRO": [
+        "Faroe Islands"
     ]
 }


app/database_etl/location_utils/country_alternative_names.json
```diff
@@ -26,5 +26,22 @@
...
+    ],
+    "GBR": [
+        "The United Kingdom",
+        "United Kingdom"
+    ],
+    "IRN": [
+        "Iran (Islamic Republic of)",
+        "Iran"
+    ], 
+    "JEY": [
```
abeljoseph, line 38: I originally intended for this file to contain one to many mappings (one ISO code maps to several country names), so I'm wondering why 1:1 mappings are needed.  

Note: the reverse mapping (country name –> ISO code) for _JEY_, _GUF_ and _FRO_ are captured in `country_iso3.json`.

rahularoradfs: Abel: see the ticket at https://airtable.com/tbli2lWQHAqBa6ZcI/viw8Ro4zYPYcBGXRw/recPEw7IDgiOWalqQ?blocks=hide. This function is in prep for that ticket.

The idea is that this ticket will identify countries in our data that aren't properly mapping to the JHU data that we need to add to this file. However, there are some countries we will never be able to actually map -- because e.g., JHU doesn't have Jersey as a country. Having these listed in this mapping means that our system won't yell at us each time it's not able to find data for Jersey.

abeljoseph (approved):
LGTM

PR #209, opened 2021-05-03
Filter out records in WHO disputed areas

atatmaja:

- Filters out records in WHO disputed areas

- https://airtable.com/tbli2lWQHAqBa6ZcI/viwWvxb27XGuziw4c/rec102UyEZJDloPx7?blocks=bipIkdtYHmboUfJut

- Ran full stack locally and manually QA'd
- Looked in local database for records that the etl found to be in disputed areas

- Must run ETL as soon as this is deployed

## Notes:

Currently we check if a circle of radius 0.5 degrees lat/lng that is centered around an estimate's coordinates intersects with a disputed region. We do this to account for pin jittering. However this is not ideal because it leads to us filtering quite a few estimates that shouldn't be filtered out (particularly estimates around the Great Lakes), and is quite slow (checking if a single coordinate is within a disputed region would probably run faster). 

In order to fix this I think we should consider using a different pin jittering method. IIRC @brett-stephen mentioned that we could take a client side approach, maybe something worth looking into before we merge this.

EDIT: There aren't any pins that are so close to contentious disputed areas that there is a risk of them falling in them due to jitter, so I am now just using the coordinate itself to check rather than a circle. However, we consider using mapbox's clustering instead of our pin jittering method soon.


rahularoradfs (commented):
Overall, all of this very much checks out to me. I think computing latlng, then checking disputed, then finding some solution to clustering (either jittering or soon via Mapbox) is the way to go for sure. Read all files, didn't have any particular comments. A granular review from our backend guru or GIS expert could be helpful!

app/database_etl/location_utils/location_util_functions.py
```diff
@@ -76,4 +81,66 @@ def get_city(row):
...
+        return in_disputed_area
+
+    # apply row_in_disputed_area across the whole df
+    df['in_disputed_area'] = df.apply(lambda row: row_in_disputed_area(row), axis=1)
+    return df
+
+
+# Get pin latitude and longitude for a record
+# Param geo_dfs = dictionary of city, state, and country DFs
+def get_record_coordinates(record: pd.Series, geo_dfs: dict) -> Tuple:
```
simonarocco, line 113: can we just define geo_dfs in here? why are we passing it in all the way from etl_main

app/database_etl/location_utils/location_util_functions.py
```diff
@@ -76,4 +81,66 @@ def get_city(row):
...
+# Checks whether each row of an input city or state dataframe
+# is in one of the WHO's disputed areas
+def check_if_in_disputed_area(df: pd.DataFrame) -> pd.DataFrame:
+    WHO_FL_URL = "https://services.arcgis.com/5T5nSi527N4F7luB/arcgis/rest/services/DISPUTED_AREAS_mask/FeatureServer/0"
+    # Create feature layer object
+    disputed_areas_fl = FeatureLayer(WHO_FL_URL)
+
+    # Defining this as a closure so that we have access to
+    # disputed_areas_fl
+    def row_in_disputed_area(row: pd.Series) -> bool:
```
simonarocco, line 95: I know why you defined it as a closure but I'm not rly a huge fan of this pattern tbh. You could just pass in disputed areas fl like you did below when you passed in the geo_dfs

simonarocco (commented):
Minor suggestions. otherwise fine as long as we've tested with and without the disputed regions flag (i.e. both true and false)

app/database_etl/location_utils/location_util_functions.py
atatmaja: geo_dfs contains the dfs for the city, state, and country tables. We don't have access to those tables here, hence we must pass them from etl_main

atatmaja:
@rahularoradfs I will be pushing this late tonight because these changes will break serotracker.com before the ETL is run

PR #211, opened 2021-05-07
Add HRP and Income Classification Columns

atatmaja:

Adds HRP and Income classification columns to country table and /records endpoint

https://airtable.com/tbli2lWQHAqBa6ZcI/viwWvxb27XGuziw4c/recTzL9dF6hFD0QyN?blocks=hide

- Tested endpoint locally to ensure that HRP and income classification fields are shown correctly
- Regression tested front end pointing to local server

N/A


app/database_etl/postgres_tables_handler/table_generator.py
```diff
@@ -203,6 +204,24 @@ def create_country_df(dashboard_source_df, current_time):
     country_df = add_latlng_to_df("country", "country_name", country_df)
     country_df['country_iso3'] = country_df["country_name"].map(lambda a: get_country_code(a))
 
+    # Populate HRP status column
```
rahularoradfs, line 207: Thanks! To be clear, even though these are in the Countries table, they will be exposed on a standard data request, right?

rahularoradfs: It seems like the GBD region and WHO region are also attributes of a country. Is there a reason that these fields aren't also in the country table, or that we are dealing with these fields differently?

app/database_etl/postgres_tables_handler/table_generator.py
```diff
@@ -203,6 +204,24 @@ def create_country_df(dashboard_source_df, current_time):
...
+    hrp_csv_path = os.path.dirname(os.path.abspath(__file__)) + '/country_hrp_map.csv'
+    hrp_countries = set(pd.read_csv(hrp_csv_path)['country_iso_code'])
+    country_df['hrp_class'] = country_df['country_iso3'].map(lambda a: 'HRP' if a in hrp_countries else 'non-HRP')
+
+    # Populate income status column
+    income_csv_path = os.path.dirname(os.path.abspath(__file__)) + '/country_income_class_map.csv'
+    income_class_df = pd.read_csv(income_csv_path)
+    # defining as a closure so that we have access to income_class_df
```
rahularoradfs, line 215: What is best practice in a case like this -- defining as a closure, or passing in the mapping df? Asking to improve my understanding of best practices.

rahularoradfs (approved):
Thanks for the ultra-fast turnaround on this! The PR itself looks great. The comments I have are more so re: overall code architecture and my own learning.

app/database_etl/postgres_tables_handler/table_generator.py
atatmaja: I think using a closure may be slightly more efficient here (though probably not noticeable) because we wouldn't have to pass the df around in memory so much.

I would consider not using a closure if the function in question could be called in more than 1 context, but that doesn't seem like it will be the case here.

@simonarocco thoughts?

app/database_etl/postgres_tables_handler/table_generator.py
atatmaja: Not sure, IMO they should be in the country table as well

rahularoradfs:
Sounds great to me!

On May 8, 2021, simonarocco ***@***.***> wrote:
> > Maybe something to ticket as a backend improvement?
> >  On May 8, 2021, simonarocco @.***> wrote: @atatmaja
> > <https://github.com/atatmaja> commented on this pull request. In
> > app/database_etl/postgres_tables_handler/table_generator.py
> > <<https://github.com/serotracker/iit->
> > backend/pull/211#discussion_r628781769>: @@ -203,6 +204,24 @@ def
> > create_country_df(dashboard_source_df,
> >  current_time): country_df = add_latlng_to_df("country",
> > "country_name", country_df) country_df['country_iso3'] =
> > country_df["country_name"].map(lambda a: get_country_code(a)) + #
> > Populate HRP status column
> >  Not sure, IMO they should be in the country table as well
> >  — You are receiving this because your review was requested. Reply
> > to this email directly, view it on GitHub
> > <<https://github.com/serotracker/iit->
> > backend/pull/211#discussion_r628781769>, or unsubscribe
> > <<https://github.com/notifications/unsubscribe->
> > auth/AOZY6J7ELCIQND6KFARWRR3TMV7Z3ANCNFSM44KVYLNQ>.
> >
>
> Something to give to @annoyingpr <https://github.com/annoyingpr> to
> ease him into BE dev?
>

—
You are receiving this because your review was requested.
<https://github.com/serotracker/iit-backend/pull/211#issuecomment-
835473641>, or unsubscribe
auth/AOZY6JZFEHXPUJ7PHGRS3JTTMWDCFANCNFSM44KVYLNQ>.

app/database_etl/postgres_tables_handler/table_generator.py
simonarocco: I mentioned this in my other PR review -- I personally don't like the pattern of using closures. I think it makes the code less readable and if we start arbitrarily defining some functions as closures while others not as closures (i.e. if diff people write the code with different styles) it can create inconsistencies which will make it harder to maintain. I don't think we will see any noticeable effects of passing the df around.

simonarocco (commented):
good but pref to not use closures for reasons listed

PR #213, opened 2021-05-14
Subgeography estimates

atatmaja:

Add parameter that dictates whether or not to serve subgeography estimates.

https://airtable.com/tbli2lWQHAqBa6ZcI/viwWvxb27XGuziw4c/recfewIYoRxjM5kSF?blocks=hide

- Tested /filter_options endpoint and /records endpoint (with include_subgeography_estimates = true and false) using postman
- Ran serotracker.com client locally pointing to local server, smoke tested all pages

Migration


app/utils/estimate_prioritization/estimate_prioritization.py
```diff
@@ -121,6 +123,28 @@ def get_prioritized_estimates(estimates: pd.DataFrame,
...
+
+    if pool:
+        selected_estimate_df = pd.concat(selected_estimates, axis=1).T.astype(estimates.dtypes.to_dict())
+    else:
+        selected_estimate_df = pd.concat(selected_estimates)
+
+    return selected_estimate_df
+
+
+# Gets prioritized estimates without pooling by a specified subgrouping variable
```
rahularoradfs, line 135: Could be helpful here to specify that this also returns prioritized estimates that are pooled -- i.e., an overall estimate, as well as the subgroups by that variable. The function name doesn't really make that clear.

app/utils/estimate_prioritization/estimate_prioritization.py
```diff
@@ -121,6 +123,28 @@ def get_prioritized_estimates(estimates: pd.DataFrame,
...
+# "subgroup_var" = subgrouping variable to refrain from pooling by
+def get_prioritized_estimates_without_pooling(estimates: pd.DataFrame,
+                                subgroup_var: str,
+                                filters: Union[tuple, None] = None,
+                                **kwargs):
+    normal_estimates = get_prioritized_estimates(estimates, filters = filters, **kwargs)
+    unpooled_filters = set(filters) if filters else set()
+    unpooled_filters = tuple(set.union(unpooled_filters,
+                                       set([lambda estimate: estimate['subgroup_var'] == subgroup_var])))
```
rahularoradfs, line 144: do we want the variable to only be subgroup_var? I wonder whether we want to make this a bit more general. Can't think of a specific use case off the bat -- most of the time it is subgroup var -- but...

app/utils/estimate_prioritization/estimate_prioritization.py
```diff
@@ -121,6 +123,28 @@ def get_prioritized_estimates(estimates: pd.DataFrame,
             selected_estimates.append(get_pooled_estimate(study_estimates))
         else:
             selected_estimates.append(study_estimates)
-    
-    selected_estimate_df = pd.concat(selected_estimates, axis = 1).T.astype(estimates.dtypes.to_dict())
-    return selected_estimate_df
\ No newline at end of file
+
+    if pool:
```
rahularoradfs, line 127: The other way to do this would be to group by the thing you don't want to pool by as part of the estimate prioritization step. You'd need to pass in a second variable to group by:

for study_name, study_estimates in estimates.groupby('study_name'):
  for subgroup_name, subgroup_estimates in subgroup_estimates.groupby(subgroup_var):

and, in this case, subgroup_var could be something like province

tbh, I think this is a bit of a worse solution, because to implement this for geography would require using a subgroup var that is concatenated country and city and state...

yeah this is a bad idea i should go to bed but imma just leave this comment anyway

app/utils/get_filtered_records.py
```diff
@@ -129,7 +128,7 @@ def get_all_records(research_fields=False, include_disputed_regions=False):
 
 
 def get_filtered_records(research_fields=False, filters=None, columns=None, include_disputed_regions=False,
-                         sampling_start_date=None, sampling_end_date=None,
+                         sampling_start_date=None, sampling_end_date=None, include_subgeography_estimates=False,
```
rahularoradfs, line 131: Is this how we want to expose this? This isn't very general. I would prefer something like:

pool = False
subgroup_estimates_by = ('subgroup_var', 'Geographical area')
include_primary_estimate_when_pooling = True

rahularoradfs (commented):
Overall looks good. Mostly reviewed for overall approach... detailed BE review on this one I trust Simona for!

In a few places, this functionality is written in a way that is uber-specific for this use case, and I wonder whether it's better to write it to be a bit more general:

- to allow subgroup estimates by one variable without also adding the primary estimate across that variable for each study
- to enable this functionality for more subgroup_var options than just geographical area
- etc...

Mostly there!

app/utils/estimate_prioritization/estimate_prioritization.py
atatmaja: I get that we want generality here, just thought that the original name proposed "donotpoolby" would be hard to interpret and confusing for future devs. I'll try to think of better nomenclature

app/utils/get_filtered_records.py
atatmaja: Depends how much control we want to allow our users to have. With more control comes more confusion and more room for error. I would rather design for what we know we will need and what we know people will be able to use properly.

simonarocco (approved):
g

PR #215, opened 2021-05-15
Disputed area pins retry logic

atatmaja:

Add logic to retry arcgis queries with exp backoff to prevent ETL failures, and log information if max retry counts are exceeded

https://airtable.com/tbli2lWQHAqBa6ZcI/viwWvxb27XGuziw4c/recy9U6VajfFDLTYA?blocks=hide

Ran ETL locally and inspected data

No.


app/database_etl/location_utils/location_util_functions.py
```diff
@@ -94,7 +96,28 @@ def row_in_feature_layer(row: pd.Series, feature_layer: FeatureLayer) -> bool:
...
+            in_disputed_area = len(feature_layer.query(geometry_filter=pin_filter).features) > 0
+            continue_query = False
+        except Exception as e:
+            # send slack message if we exceed retry count
+            if retries > MAX_RETRIES:
+                body = f'Unable to check if the record with ID {row["source_id"]} is in a disputed region.'
+                send_slack_message(body, channel='#dev-logging-etl')
+                continue_query = False
+            else:
+                sleep(1.5**(retries))
```
rahularoradfs, line 118: Cognizant that this may slow down ETL considerably if we get a series of these errors. This permits 1 minute per record -- which would mean the whole ETL would take over 100+ hours to run if all records error. Thoughts on the rate of backoff / bounds of backoff here?

rahularoradfs (approved):
generally looks good, comment about speed of backoff :)

app/database_etl/location_utils/location_util_functions.py
atatmaja: Yeah I'm worried about speed as well - I put relatively high backoff to be a bit more conservative because I wasn't sure how long the arcgis server would need to recover in the event of a failure. I can try using smaller backoff (maybe like 1.2**retries) instead and then adjust if we continue to get less reliable etl runs.

rahularoradfs:
Other way to be more conservative is to reduce max n from 10 >> 5 or
similar!


On May 16, 2021, simonarocco ***@***.***> wrote:
> @atatmaja commented on this pull request.
>

> In app/database_etl/location_utils/location_util_functions.py
> <https://github.com/serotracker/iit-
> backend/pull/215#discussion_r633100872>:
>

> + # Default to setting in_disputed_area = True to ensure we never show
pins in disputed area + in_disputed_area = True + # Make query to
determine whether or not the pin is in the disputed area + # If the
query times out, retry with exponential backoff + while continue_query:
+ try: + in_disputed_area =
len(feature_layer.query(geometry_filter=pin_filter).features) > 0 +
continue_query = False + except Exception as e: + # send slack message
if we exceed retry count + if retries > MAX_RETRIES: + body = f'Unable
to check if the record with ID {row["source_id"]} is in a disputed
region.' + send_slack_message(body, channel='#dev-logging-etl') +
continue_query = False + else: + sleep(1.5**(retries))
> Yeah I'm worried about speed as well - I put relatively high backoff
> to be a bit more conservative because I wasn't sure how long the
> arcgis server would need to recover in the event of a failure. I can
> try using smaller backoff (maybe like 1.2**retries) instead and then
> adjust if we continue to get less reliable etl runs.
>

—
You are receiving this because your review was requested.
<https://github.com/serotracker/iit-
backend/pull/215#discussion_r633100872>, or unsubscribe
auth/AOZY6J5MN3B3YIMA2GY42PTTN7MXTANCNFSM446IQUUA>.

app/database_etl/location_utils/location_util_functions.py
```diff
@@ -94,7 +96,28 @@ def row_in_feature_layer(row: pd.Series, feature_layer: FeatureLayer) -> bool:
...
+    MAX_RETRIES = 9
+    # Default to setting in_disputed_area = True to ensure we never show pins in disputed area
+    in_disputed_area = True
+    # Make query to determine whether or not the pin is in the disputed area
+    # If the query times out, retry with exponential backoff
+    while continue_query:
+        try:
+            in_disputed_area = len(feature_layer.query(geometry_filter=pin_filter).features) > 0
+            continue_query = False
+        except Exception as e:
```
simonarocco, line 111: 1) remove as e (unless we are gonna use e) 2) can you add the more specific type of exception instead of just Exception?

simonarocco (commented):
overall looks good, minor comments

rahularoradfs:
@atatmaja shall we merge this to get ETL going again?

PR #219, opened 2021-05-18
add distinct to aggregate function

simonarocco:
- Apply distinct function to aggregate query function so that only distinct values are returned in the list

- Tested /records endpoint and /records/sourceid endpoints

- No


atatmaja (approved):
LGTM

PR #221, opened 2021-05-21
fixing migration issue and adding part to update canadian data

simonarocco:

### DIFF
--- a/app/database_etl/tableau_data_connector/table_generator.py
+++ b/app/database_etl/tableau_data_connector/table_generator.py
@@ -14,7 +14,8 @@ def upload_analyze_csv(canadian_data):
     # Get records
     prioritize_estimates = True if canadian_data else False
     include_subgeography_estimates = True if canadian_data else False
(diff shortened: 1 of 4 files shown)


app/database_etl/tableau_data_connector/table_generator.py
```diff
@@ -10,10 +10,13 @@
 load_dotenv()
 
 
-def upload_analyze_csv():
+def upload_analyze_csv(canadian_data):
```
atatmaja, line 13: could probably find a cleaner/more generic way to do this but given that you're rushing rn I'll let you leave this as a todo

app/database_etl/tableau_data_connector/table_generator.py
```diff
@@ -10,10 +10,13 @@
...
     # Get records
+    prioritize_estimates = True if canadian_data else False
```
atatmaja, line 15: also should we have filters: {"country": ["Canada"]} if canadian_data is true?

atatmaja (approved):

PR #224, opened 2021-05-28
Endpoint testing

atatmaja:

- Adds helper functions to help populate a testing database with a realistic set of records
- Adds several basic endpoint tests for each endpoint that's still used regularly in production
- Fixes CI so that errors are shown in github actions and a postgres db is spun up and torn down with each test run

image

https://airtable.com/tbli2lWQHAqBa6ZcI/viwWvxb27XGuziw4c/recSdsSv04EjFIUsL?blocks=bipIkdtYHmboUfJut
https://airtable.com/tbli2lWQHAqBa6ZcI/viwWvxb27XGuziw4c/recHjpOfvZUH0SSHT?blocks=bipIkdtYHmboUfJut

See github actions CI results!

N/A


.github/workflows/ci.yml
```diff
@@ -9,25 +9,26 @@ jobs:
...
+        postgresql password: 'postgres'
     - name: Setup Python
       uses: actions/setup-python@v2
       with:
         python-version: '3.8.3'
     - name: Install
       run: |
         pip install -r requirements.txt
-        mkdir data/
-        curl "https://raw.githubusercontent.com/owid/covid-19-data/master/public/data/vaccinations/vaccinations.csv" > data/vaccinations.csv
```
rahularoradfs, line 20: Do we not need this as part of CI anymore? Or is this more of a CD thing? Bit confused

app/__init__.py
```diff
@@ -4,8 +4,8 @@
 from flask import Flask
 from flask_restplus import Api
 from flask_cors import CORS
-from flask_sqlalchemy import SQLAlchemy
 from .config import config_by_name
+from flask_sqlalchemy import SQLAlchemy
```
rahularoradfs, line 8: solid change tx

app/config.py
```diff
@@ -38,17 +38,14 @@ def __init__(self):
 
 
 class ApiTestingConfig(ApiConfig):
-    DATABASE_USERNAME = os.getenv('DATABASE_USERNAME')
-    DATABASE_PASSWORD = os.getenv('DATABASE_PASSWORD')
+    DATABASE_USERNAME = os.getenv('DATABASE_USERNAME', 'postgres')
```
rahularoradfs, line 41: are there any security issues here with using postgres/postgres as user/pass? probably not, right, as this is only as part of CI/CD testing?

app/serotracker_sqlalchemy/db_model_config.py
```diff
@@ -27,13 +27,17 @@
...
              'city_bridge',
              'state',
              'state_bridge',
              'test_manufacturer',
-             'test_manufacturer_bridge']
+             'test_manufacturer_bridge',
+             'antibody_target',
+             'antibody_target_bridge']
 
 # TODO: instead of using this config, programmatically generate a list of columns for the DashboardSource table
```
rahularoradfs, line 42: this is prob part of schema refactor

app/serotracker_sqlalchemy/db_model_config.py
```diff
@@ -27,13 +27,17 @@
     ]
 }
 
-db_tables = ['airtable_source',
+db_tables = ['dashboard_source',
```
rahularoradfs, line 30: bit confused -- how was the app working if these were not part of the db model config before?

tests/app_setup/test_app_setup.py
```diff
@@ -1,71 +1,10 @@
 import os
 
 
-def test_flask_environment_variable():
+def test_app_setup(app):
```
rahularoradfs, line 4: this feels like a less descriptive name to me!

tests/factories.py
```diff
@@ -1,180 +1,245 @@
...
-            sqlalchemy_session_persistence = 'commit'
-        id = uuid.uuid4()
-        source_id = uuid.uuid4()
-        test_manufacturer_id = uuid.uuid4()
-        created_at = factory.LazyFunction(datetime.now)
-    return TestManufacturerBridgeFactory(**kwargs)
+# Note: Using this pattern instead of the previous function based pattern
+# because we need to use the same class throughout our test fixture so that
+# "sequence functions" work properly
+class DashboardSourceFactory(factory.alchemy.SQLAlchemyModelFactory):
```
rahularoradfs, line 15: Bit concerned with this pattern for Factories for testing. This means that we need to add to and delete from these factory files each time we modify our schema files. That means having aspects of our database schema in multiple places. Any way to get around this? 

There's a part of me that was thinking about having one definitive CSV-based schema file. This could have columns that are something like: 

- variable name in airtable
- which airtable table the variable is in 
- corresponding variable name in database
- datatype in database 
- table in database (you could use these three vars to programatically generate the models.py file - metaprogramming)
- any data cleaning that needs to happen in the ETL (e.g., unwrapping list columns, etc...)
- you could add a column re what function should be used for testing: LazyFunction? Sequence? etc

tests/factories.py
```diff
@@ -1,180 +1,245 @@
...
+    class Meta:
+        model = DashboardSource
+        sqlalchemy_session = db.session
+        sqlalchemy_session_persistence = 'commit'
+
+    source_id = factory.LazyFunction(uuid.uuid4)
```
rahularoradfs, line 21: It's going to take some pretty careful design of these factories to ensure that our endpoints are testing all functionality. Is this desirable, or not so much?

tests/utils/helper_funcs.py
```diff
@@ -8,68 +11,39 @@ def delete_records(session, records):
...
     test_manufacturer_id = new_test_manufacturer_record.test_manufacturer_id
 
-    # Insert joined record into test manufacturer table
-    factories.test_manufacturer_factory(session, test_manufacturer_id=test_manufacturer_id)
+    # Insert records into dashboard_source, research_source, and bridge tables
+    for i in range(num_records):
+        source_id = uuid.uuid4()
+        # Ensure all are dashboard_primary estimates so that estimate prioritization
+        # always returns 1 estimate per study
+        factories.DashboardSourceFactory(source_id=source_id, country_id=country_id, study_name=study_name)
```
rahularoradfs, line 42: again, realism here -- what would a realistic set of estimates look like? prob 1x dash primary, 1x academic primary, then 10x random other estimates with varying other things... not sure how important this is for the sort of testing we're doing, though!

tests/app_setup/test_db_setup.py
```diff
@@ -1,42 +1,58 @@
...
-def test_attached_db(db):
-    assert isinstance(db, SQLAlchemy)
-
-
-def test_db_table_names():
-    engine = create_engine(app.config['SQLALCHEMY_DATABASE_URI'])
+def test_db_table_names(db):
     for table in db_tables:
-        assert table in engine.table_names()
+        assert table in db.engine.table_names()
```
rahularoradfs, line 11: Don't we also need to test this the other way?

You could assert that the two table names are identical // equivalent sets, or otherwise assert that every table in db.engine.table_names() is in db_tables to test it the other way

tests/app_setup/test_db_setup.py
```diff
@@ -1,42 +1,58 @@
...
-        assert len(_session.query(AntibodyTargetBridge).all()) == 0
-        assert len(_session.query(City).all()) == 0
-        assert len(_session.query(CityBridge).all()) == 0
-        assert len(_session.query(State).all()) == 0
-        assert len(_session.query(StateBridge).all()) == 0
-        assert len(_session.query(TestManufacturer).all()) == 0
-        assert len(_session.query(TestManufacturerBridge).all()) == 0
-
-
+    assert session.query(DashboardSource).count() == 0
```
rahularoradfs, line 15: Wouldn't this be better done iterating over the list of tables in the database? This is fragile because it breaks if we add or remove a table

tests/app_setup/test_db_setup.py
```diff
@@ -1,42 +1,58 @@
...
+    all_dashboard_source_records = session.query(DashboardSource).all()
+    assert [new_dashboard_source_record] == all_dashboard_source_records
+    assert len(all_dashboard_source_records) == 1
+    delete_records(session, all_dashboard_source_records)
+    assert session.query(DashboardSource).count() == 0
+
+
+# Tests inserting a complete record set
+# using our test helper functions
+def test_insert_recordset(session):
```
rahularoradfs, line 41: same comment as above re this being fragile if we add or remove tables

tests/endpoints/test_endpoints.py
```diff
@@ -0,0 +1,107 @@
...
+    assert data == 'The healthcheck endpoint was hit.'
+
+
+# Test get filter options
+def test_get_filter_options(client):
+    response = client.get('/data_provider/filter_options')
+    status_code = response.status_code
+    data = response.get_json()
+    assert status_code == 200
+    assert type(data) is dict
```
rahularoradfs, line 20: any other assertions re the return that we should be making here? or does that assume that the form of filter_options doesn't change?

tests/endpoints/test_endpoints.py
```diff
@@ -0,0 +1,107 @@
...
+    response = client.get('/healthcheck/')
+    status_code = response.status_code
+    data = response.get_json()
+    assert status_code == 200
+    assert type(data) is str
+    assert data == 'The healthcheck endpoint was hit.'
+
+
+# Test get filter options
+def test_get_filter_options(client):
```
rahularoradfs, line 15: for all of these tests, is there a way to store them in the same place as the endpoint so that when the endpoint is modified it is natural to modify these too?

tests/endpoints/test_endpoints.py
rahularoradfs: thinking about this across the rest of the endpoints too - like, is there a way to make sure we're not just getting a janky empty dict back? or si that not the purpose of these things to test?

tests/endpoints/test_endpoints.py
```diff
@@ -0,0 +1,107 @@
...
+            "country": ["country_name_1"]
+        }
+    })
+    status_code = response.status_code
+    data = response.get_json()
+    assert status_code == 200
+    assert len(data) == 1
+
+
+# Test get records with an empty request body
```
rahularoradfs, line 56: these get_records tests are great -- thanks!!

rahularoradfs (commented):
Looks great, @atatmaja!!

Left a few more structural // strategic comments, focusing on test maintainability more than anything. Overall, this PR is very very much a net positive.

Would still love to see a design doc describing our overall testing strategy. How are we leveraging these sorts of 'overall tests'? Is there a role for unit testing and if so, where? What about integration testing? What about testing user features (thinking about the Airtable data download here)? What other tests do we want to build? Per our lengthy discussion on Slack that evening.

Not approving because this PR definitely merits more granular code review by @simonarocco :)

app/config.py
atatmaja: Nah

app/serotracker_sqlalchemy/db_model_config.py
atatmaja: Bit misleading but this object is only being used in the test setup, could probably do without this tbh

tests/app_setup/test_db_setup.py
atatmaja: Yeah we could do that. Also my suspicion is that there should be a way to infer all tables from a models file so we shouldn't need db_tables anyways

tests/app_setup/test_db_setup.py
atatmaja: Agreed, can take an approach similar to what I described above

tests/endpoints/test_endpoints.py
atatmaja: We could, I would need to modify our test runner but that can be arranged

tests/endpoints/test_endpoints.py
atatmaja: we could store a list of keys that filter options should have and assert that `data` contains those.

.github/workflows/ci.yml
atatmaja: We changed how we did this a while ago to always download the OWID CSVs on each ETL run instead of relying on a local file - so we don't need this step in CI or CD atm

app/serotracker_sqlalchemy/db_model_config.py
rahularoradfs: yes, i think just best to read from the list of tables and not have a static obj

tests/endpoints/test_endpoints.py
rahularoradfs: that's possible -- I then become worried about that being a fragile test if there is a change in our data. Not sure ow to straddle this line.

tests/endpoints/test_endpoints.py
rahularoradfs: Think this is worthwhile!

tests/endpoints/test_endpoints.py
atatmaja: Eventually filter_options endpoint will pull data from airtable - I think once that happens it should be easier to design a more robust test. 

I'm more inclined to leave this as is for now so that we don't spend time overengineering a test

app/utils/estimate_prioritization/estimate_prioritization_test.py
```diff
@@ -3,24 +3,27 @@
 import pandas as pd
 import json
 
+# TODO: Update these to reflect current state of est prio and use our unit testing infrastructure
```
atatmaja, line 6: Happy birthday @rahularoradfs https://airtable.com/tbli2lWQHAqBa6ZcI/viw7Tba8QVTJ49F4v/recAO1E99BJXnTG8s?blocks=bipIkdtYHmboUfJut

app/namespaces/data_provider/test_data_provider_endpoints.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+# Test get records details
+def test_get_record_details(client):
+    source_id = str(_db.session.query(DashboardSource.source_id).all()[0][0])
+    response = client.get(f'/data_provider/record_details/{source_id}')
+    status_code = response.status_code
+    data = response.get_json()
+    assert status_code == 200
+    assert type(data) is dict
```
simonarocco, line 30: do we wanna do any more granular asserts where we check that the source_id is actually contained as a key in the result (i.e. so we confirm it got the correct record corresponding to the desired source id)?

app/namespaces/data_provider/test_data_provider_endpoints.py
```diff
@@ -0,0 +1,107 @@
+from app import db as _db
```
simonarocco, line 1: OVERALL COMMENT: I don't like that all these files have been taken out of the /tests folder. It's best practice IMO to keep the testing stuff in a test folder. Can we plz move it back to the original structure?

app/namespaces/data_provider/test_data_provider_endpoints.py
```diff
@@ -0,0 +1,107 @@
...
+	    "filters": {},
+        "research_fields": True
+    })
+    status_code = response.status_code
+    data = response.get_json()
+    assert status_code == 200
+    assert len(data) == 4
+    # check if some field in the research table is in the returned record
+    for record in data:
+        assert "adj_sensitivity" in record
```
simonarocco, line 107: good

test_utils/helper_funcs.py
```diff
@@ -0,0 +1,49 @@
+import test_utils.factories as factories
```
simonarocco, line 1: AGAIN: I think /test_utils should be within a greater directory called /tests

test_utils/setup_tests/test_app_setup.py
```diff
@@ -0,0 +1,10 @@
+import os
```
simonarocco, line 1: Again, I think we need to change the structure. This file is a test, not a test util (a test util in my mind is a helper function, not an actual test), so we should put this file in /tests and have /test_utils be within that folder

simonarocco (commented):
overall good, thanks for adding more test coverage, but I think we should reconsider the structure of the tests in our repo. I think we should have /tests and /test_utils inside tests which should only contain helper functions. actual test files should go in /tests. also make sure that we follow the naming convention of all test files where they start with test_ (calling out the estimate prio file)

app/namespaces/data_provider/test_data_provider_endpoints.py
atatmaja: I actually kind of agree with Rahul that having tests in the same directory as the codebase being tested will make people more likely to update tests, so I'm not entirely in favour of this.

@simonarocco @rahularoradfs

app/namespaces/data_provider/test_data_provider_endpoints.py
atatmaja: whoops missed that

atatmaja: I don't think this endpoint actually returns source_id currently

rahularoradfs (approved):

PR #226, opened 2021-06-05
Fix null pins problem

atatmaja:

ETL was computing null pin_latitudes/longitudes when it shouldn't have been. Previously, we would first determine a location or set of locations that would represent a record and then compute record's latlng based on those locations. The issue with that strategy is that we may not have a latlng for that location. Therefore, this PR amends that by tweaking the logic to compute fallback locations to attribute the record in hopes of always calculating a latlng for a given record.

https://airtable.com/tbli2lWQHAqBa6ZcI/viwWvxb27XGuziw4c/recSAReGyBHSNawta?blocks=bipIkdtYHmboUfJut
- note there may also be an issue with the country 

Ran ETL, ensured that there were no records with null latlngs (unless the record is somehow not affiliated with a country).

No


app/database_etl/etl_main.py
```diff
@@ -89,12 +89,15 @@ def main():
         research_source, research_source_cols = create_research_source_df(dashboard_source)
 
         # Compute pin information for each record in dashboard source table
+        # modify dashboard source structure so country col is similar to city and state col
+        modified_dashboard_source = dashboard_source
+        modified_dashboard_source['country'] = modified_dashboard_source['country'].apply(lambda x: [x])
```
rahularoradfs, line 94: Is this best done here, or with a type check later on to see whether the item in question is a list or str?

app/database_etl/location_utils/location_util_functions.py
```diff
@@ -124,29 +124,29 @@ def row_in_feature_layer(row: pd.Series, feature_layer: FeatureLayer) -> bool:
...
         if record[region_type] and len(record[region_type]) > 0:
             pin_regions = record[region_type]
-            pin_region_type = region_type
+            geo_df = geo_dfs[region_type]
+            col_name = f"{region_type}_name"
+            pin_lats = [geo_df[geo_df[col_name] == region_name].iloc[0]['latitude'] for region_name
+                        in pin_regions if pd.notnull(geo_df[geo_df[col_name] == region_name].iloc[0]['latitude'])]
+            pin_lngs = [geo_df[geo_df[col_name] == region_name].iloc[0]['longitude'] for region_name
+                        in pin_regions if pd.notnull(geo_df[geo_df[col_name] == region_name].iloc[0]['longitude'])]
+            if len(pin_lats) > 0:
```
rahularoradfs, line 139: There is an edge case here where pin_lats and pin_longs don't have the same length that it might be worth addressing. The code here assumes that latitude is NA iff longitude is NA

app/database_etl/location_utils/location_util_functions.py
```diff
@@ -124,29 +124,29 @@ def row_in_feature_layer(row: pd.Series, feature_layer: FeatureLayer) -> bool:
 # Get pin latitude and longitude for a record
 # Param geo_dfs = dictionary of city, state, and country DFs
 def get_record_coordinates(record: pd.Series, geo_dfs: dict) -> Tuple:
-    pin_regions = [record['country']]
-    pin_region_type = 'country' if record['country'] is not None else ''
+    pin_lat = None
+    pin_lng = None
 
-    for region_type in ['state', 'city']:
+    for region_type in ['country', 'state', 'city']:
```
rahularoradfs, line 130: I'm a bit confused by the design pattern here. 

What I might have expected this to look like is something like the following. In essence, this code looks from small to large to see whether it can find a latlong for the most granular geography type, and if not, steps back one broader.

```
for region_type in ['city', 'state', 'country']: # from small to large
  if not record[region_type]:
    continue
  lat, lng = get_latlong(set(record[region_type], record['country']))
  # using the two in combination to ensure that you don't get e.g., Calgary in Scotland instead of Calgary in Canada
  # with some sort of deduplication if region_type is country
  if lat and lng: 
    break
```

app/database_etl/location_utils/location_util_functions.py
```diff
@@ -124,29 +124,29 @@ def row_in_feature_layer(row: pd.Series, feature_layer: FeatureLayer) -> bool:
...
         if record[region_type] and len(record[region_type]) > 0:
             pin_regions = record[region_type]
-            pin_region_type = region_type
+            geo_df = geo_dfs[region_type]
+            col_name = f"{region_type}_name"
+            pin_lats = [geo_df[geo_df[col_name] == region_name].iloc[0]['latitude'] for region_name
```
simonarocco, line 135: can we store geo_df[geo_df[col_name] == region_name].iloc[0]['latitude'] as a sep variable that we define first just for readability and cause it's repeated

app/database_etl/location_utils/location_util_functions.py
simonarocco: in theory yes, but in practice do you actually think there is a case where there is a lat and not a lng? I don't know about that

simonarocco: i guess you could just add and len(pin_lngs) > 0

app/database_etl/location_utils/location_util_functions.py
simonarocco: I kind of also agree with this -- I'm confused why we are averaging the pin lat lngs from the city, state and country? Wouldn't we want the most accurate lat long value if it's avail? i.e. if city is available, take city, otherwise take state, otherwise take country? maybe I don't understand the OG logic too well myself either

simonarocco (commented):
confused by OG logic. u me and rahul should sync on this maybe

app/database_etl/location_utils/location_util_functions.py
atatmaja: Instead of going from granular to least granular and backtracking if necessary, this logic goes from least granular to most granular and breaks when we can no longer go more granular. That way no backtracking is necessary (which I would find confusing)

app/database_etl/location_utils/location_util_functions.py
atatmaja: This should never happen. The api call will return both a lat and lng. I can add this extra check for good measure but confident that wouldn't happen

app/database_etl/etl_main.py
atatmaja: The purpose of this is to get the country field to be similar in structure with the city and state fields. I did this to simplify the logic in the function used to get lats and longs.

Originally the country col would always be a string, just making it a one element list for consistency

simonarocco: this makes sense to me

app/database_etl/location_utils/location_util_functions.py
atatmaja: We are averaging in the case where we have multiple cities or multiple states. We made a decision that if say we had a study conducted in multiple cities, the pin would show up in the center of those cities.

atatmaja: @rahularoradfs using a country and city to narrow down on a location isn't enough. A country can have multiple of the same cities if they're they're different states. The problem is right now our airtable schema is not robust enough to be able to infer these relationships (we simply store lists of cities and states for each record).

Why don't we consider improving this in our airtable refactor?

app/database_etl/etl_main.py
```diff
@@ -88,6 +88,18 @@ def main():
         # Create research source based on dashboard source
         research_source, research_source_cols = create_research_source_df(dashboard_source)
 
+        # remove state names from city name column and place them in their own column
+        multi_select_tables_dict["city"]["state_name"] = multi_select_tables_dict["city"]["city_name"] \
```
rahularoradfs, line 92: Why do we need to do this in the first instance? What is it about our data structure that makes this necessary? Seems a bit convoluted.

app/database_etl/location_utils/location_util_functions.py
```diff
@@ -124,28 +124,53 @@ def row_in_feature_layer(row: pd.Series, feature_layer: FeatureLayer) -> bool:
...
+    pin_lat, pin_lng = None, None
+
+    # First try to query a matching country for the record and temporarily assign
+    # its pin coordinates to it
+    if not record['country']:
+        # If there is no country associated with the record there is no way we can query a latlng, return None
+        return pin_lat, pin_lng
+    country_df = geo_dfs['country']
+    matching_country = country_df[country_df['country_name'] == record['country']].iloc[0]
+    pin_lat, pin_lng = matching_country['latitude'], matching_country['longitude']
```
rahularoradfs, line 136: what happens if matching_country is none here? Is that ever the case, or are we sure that our country db is comprehensive?

rahularoradfs (approved):
makes mucho sense to me!

app/database_etl/location_utils/location_util_functions.py
atatmaja: The country_df is constructed by taking the unique set of countries that exist in our data, matching_country should always return a record

atatmaja: We do have cases where matching_country will be incomplete tho (e.g. no ISO2 code for vatican city - which we need to update when we move to airtable)

app/database_etl/etl_main.py
atatmaja: @rahularoradfs 2 reasons:

1. Able to get better results when querying for {city},{state} as opposed to {city}
2. Needed to ensure that the city_name column was unique so that the logic we wrote to construct bridge tables would work properly (see https://github.com/serotracker/iit-backend/blob/master/app/database_etl/postgres_tables_handler/table_generator.py#L191)

Honestly this probably deserves a refactor. If our city table contained a city_name and state_name column from the start and we modified our bridge table construction logic a bit, it would be much cleaner to read. If it’s okay I’d rather do this in a separate PR so that we can push incremental changes out.

rahularoradfs: agree on intended change, agree on sep PR :) would love to merge this one ASAP as there are large user implications of this, incl mali and south sudan studies showing up

app/database_etl/location_utils/location_util_functions.py
rahularoradfs: got it!

PR #230, opened 2021-06-15
Revert "Fix null pins problem"

rahularoradfs:
Reverts serotracker/iit-backend#226


atatmaja:
taking this down cuz we realized this wasn't this issue

PR #239, opened 2021-07-22
Add airtable filter ordering for population_group

annoyingpr:
https://airtable.com/tbli2lWQHAqBa6ZcI/viwTYqtBDdt3Wt3Yo/rec5XzyfWRX4GQ3N4?blocks=hide


app/database_etl/airtable_records_handler/airtable_records_getter.py
```diff
@@ -84,3 +91,47 @@ def get_all_records():
         }
         send_api_error_slack_notif(body, data, error=e, request_info=request_info, channel='#dev-logging-etl')
         return request_info
+
+def ingest_sample_frame_goi_filter_options():
+    url = "https://api.airtable.com/v0/appFfYIupTrVQ0HJx/Sample%20Frame%20GOI?fields=Order&fields=Name&sort%5B0%5D%5Bfield%5D=Order&fields=French%20Name"
```
atatmaja, line 96: What do you think about storing these magic constants more nicely, maybe there's a constants file we can use to store all URLs

app/database_etl/airtable_records_handler/airtable_records_getter.py
```diff
@@ -84,3 +91,47 @@ def get_all_records():
...
+    headers = {'Authorization': 'Bearer {}'.format(AIRTABLE_API_KEY)}
+
+    # Make request and retrieve records in json format
+    r = requests.get(url, headers=headers)
+    data = r.json()
+
+    # Try to get records from data if the request was successful
+    try:
```
atatmaja, line 104: This request/retry logic seems pretty similar to the main Airtable request function (and I feel like we may use it more in the future) - let's consider making this a more general function

app/database_etl/airtable_records_handler/airtable_records_getter.py
```diff
@@ -84,3 +91,47 @@ def get_all_records():
...
+        # Get sorted records
+        records = data["records"]
+        drop_specific_table_entries(datetime.now(), PopulationGroupOptions)
+
+        # Add records to table 
+        with db_session() as session:
+            records_to_add = [PopulationGroupOptions(id=uuid4(), 
+                                                    name=record['fields']['Name'], 
+                                                    french_name=record['fields']['French Name'] if 'French Name' in record['fields'] else None,
```
atatmaja, line 113: Are you sure we won't need `record['fields']['Order'] if 'Order' in record['fields'] else None` as well? I don't think order will be guaranteed to be in the table

app/database_etl/airtable_records_handler/airtable_records_getter.py
```diff
@@ -84,3 +91,47 @@ def get_all_records():
...
+        return records
+    except KeyError as e:
+        body = "Results were not successfully retrieved from Airtable API. " \
+               "Please check connection parameters in config.py and fields in airtable_fields_config.json."
+        request_info = {
+            "url": url,
+            "headers": json.dumps(headers)
+        }
+        logging.error(e)
+        # send_api_error_slack_notif(body, data, error=e, request_info=request_info, channel='#dev-logging-etl')
```
atatmaja, line 128: Should uncomment??

app/namespaces/data_provider/data_provider_service.py
```diff
@@ -211,4 +212,8 @@ def get_all_filter_options() -> Dict[str, Any]:
             0].isoformat()
         options["updated_at"] = session.query(func.max(DashboardSource.publication_date))[0][0].isoformat()
 
+        # Get population group options
+        results = session.query().with_entities(PopulationGroupOptions.order, PopulationGroupOptions.name, PopulationGroupOptions.french_name)
+        options["population_group"] = [[result[1], result[2]] for result in sorted(results, key=lambda result: result[0])]
```
atatmaja, line 217: Are we sure this sorting will work properly if the order is null? I think the behaviour should be that options with null order should just be shoved at the end.

Also can we put a comment describing what we are sorting by since it is not clear on first read.

atatmaja (commented):
A few more minor comments, thanks @annoyingpr and lmk if u have any q's

app/database_etl/airtable_records_handler/airtable_records_getter.py
annoyingpr: definitely a good idea! doesn't look like there's a constants file for all URLs, but i was thinking i could add this Airtable URL to our `config.py`?

app/database_etl/airtable_records_handler/airtable_records_getter.py
annoyingpr: from my understanding, i thought order would be there since it's linked to the filter options right and we have automations running for updates?

app/namespaces/data_provider/data_provider_service.py
annoyingpr: for sure, will add that comment. this goes back to my above comment, but can Order be null? if it is, i agree with your approach and ill update, but i thought we would always have order because its linked to the filter options

simonarocco:
@annoyingpr can you please fill out the section describing what this PR does? good to leave a clear paper trail. preferable if we could do that both in the description of this PR and also in the airtable ticket.

app/database_etl/airtable_records_handler/airtable_records_getter.py
simonarocco: @annoyingpr can you add a comment to this function overall describing what it does? i.e. what sample frame goi filter options is

app/database_etl/airtable_records_handler/airtable_records_getter.py
```diff
@@ -84,3 +91,47 @@ def get_all_records():
...
+
+    # Make request and retrieve records in json format
+    r = requests.get(url, headers=headers)
+    data = r.json()
+
+    # Try to get records from data if the request was successful
+    try:
+        # Get sorted records
+        records = data["records"]
+        drop_specific_table_entries(datetime.now(), PopulationGroupOptions)
```
simonarocco, line 107: Won't this just delete all the entries in the population group options table? If all the entries have a date in the past? If so, what happens if this try statement fails after dropping all the records (e.g., fails on line 112), but doesn't replace the table with any records? Then we will have an empty table, right?

app/database_etl/airtable_records_handler/airtable_records_getter.py
simonarocco: @annoyingpr who can we check this with? who wrote the automation - christian?

app/database_etl/etl_main.py
```diff
@@ -136,6 +136,9 @@ def main():
         # Make sure that filter options are still valid
         check_filter_options(dashboard_source)
 
+        # Update ordering and translations for filter options 
+        ingest_sample_frame_goi_filter_options()
```
simonarocco, line 140: this function returns something but we aren't using the return info - let's remove what the function returns if we aren't gonna use it (i.e. just put return instead of return records)

app/database_etl/postgres_tables_handler/postgres_loader.py
```diff
@@ -81,3 +81,9 @@ def drop_table_entries(current_time: datetime, drop_old: bool = True):
                 session.query(table).filter(or_(table.created_at == current_time, table.created_at.is_(None))).delete()
             session.commit()
     return
+
+def drop_specific_table_entries(current_time: datetime, table_name):
```
simonarocco, line 85: are we gonna be using this function elsewhere? right now it just looks like we are using it in one place, and it looks like we are dropping all the records in the table anyways. i feel like we could just put the function logic where it's being called unless we are gonna use this again

app/namespaces/data_provider/data_provider_service.py
```diff
@@ -211,4 +212,8 @@ def get_all_filter_options() -> Dict[str, Any]:
             0].isoformat()
         options["updated_at"] = session.query(func.max(DashboardSource.publication_date))[0][0].isoformat()
 
+        # Get population group options
+        results = session.query().with_entities(PopulationGroupOptions.order, PopulationGroupOptions.name, PopulationGroupOptions.french_name)
```
simonarocco, line 216: can just do ```session.query(PopulationGroupOptions.order, PopulationGroupOptions.name, PopulationGroupOptions.french_name)``` instead

simonarocco (changes requested):
great work prannoy!! thanks for this. just some questions/considerations to discuss

app/database_etl/airtable_records_handler/airtable_records_getter.py
atatmaja: Prannoy is right I double checked. Either way though I wonder if it's a good idea to be fail safe and consistent with null checks here.

app/database_etl/postgres_tables_handler/postgres_loader.py
annoyingpr: i think we will! this logic will be used for all filter options in the future (population_group was just the first test)

app/database_etl/airtable_records_handler/airtable_records_getter.py
annoyingpr: yeah you're right... i can move the dropping old records to only happen if we've inserted new ones - good catch!

app/database_etl/airtable_records_handler/airtable_records_getter.py
annoyingpr: split it into 2 functions: `_airtable_get_request` and `handle_airtable_request`

app/database_etl/postgres_tables_handler/postgres_loader.py
atatmaja: nit: I would prefer this function being called `drop_old_table_entries`

atatmaja: Also couldn't we just work this into the `drop_table_entries` function that we have currently, where we use the same pattern with all other tables??? I missed this earlier but tbh don't think we really need to have a whole new function/workflow here.

annoyingpr: good catch, just added `PopulationGroupOptions` to the `table_names_dict` so now it will be cleared with the `drop_table_entries` function!

app/database_etl/airtable_records_handler/airtable_records_getter.py
```diff
@@ -56,17 +84,13 @@ def get_all_records():
...
-    r = requests.get(url, headers=headers)
-    data = r.json()
+    data = _airtable_get_request(url)
 
     # Try to get records from data if the request was successful
     try:
         # If offset was included in data, retrieve additional paginated records
         if 'offset' in list(data.keys()):
-            request_info = [url, headers]
+            request_info = [url, {'Authorization': 'Bearer {}'.format()}]
```
simonarocco, line 93: why did you remove headers? ```'Authorization': 'Bearer {}'.format()``` is not actually doing a replace because you didn't specify anything the format function

app/database_etl/airtable_records_handler/airtable_records_getter.py
```diff
@@ -56,17 +84,13 @@ def get_all_records():
     url = AIRTABLE_REQUEST_URL.format(AIRTABLE_BASE_ID)
     url = _add_fields_to_url(url)
     url += '&filterByFormula={ETL Included}=1'
-    headers = {'Authorization': 'Bearer {}'.format(AIRTABLE_API_KEY)}
-
-    # Make request and retrieve records in json format
-    r = requests.get(url, headers=headers)
-    data = r.json()
+    data = _airtable_get_request(url)
```
simonarocco, line 87: I think we should pass in headers here instead of redefining it in the airtable call and then possibly again in the error handling

app/database_etl/airtable_records_handler/airtable_records_getter.py
```diff
@@ -76,11 +100,39 @@ def get_all_records():
...
     except KeyError as e:
-        body = "Results were not successfully retrieved from Airtable API. " \
-               "Please check connection parameters in config.py and fields in airtable_fields_config.json."
-        request_info = {
-            "url": url,
-            "headers": json.dumps(headers)
-        }
-        send_api_error_slack_notif(body, data, error=e, request_info=request_info, channel='#dev-logging-etl')
-        return request_info
+        _handle_airtable_error(e, data, url)
```
simonarocco, line 103: again just pass in headers that you define before this try catch statement so you just define it once

app/database_etl/airtable_records_handler/airtable_records_getter.py
```diff
@@ -76,11 +100,39 @@ def get_all_records():
...
+
+'''
+Updates population_group_options table in whiteclaw with entries from Sample Frame GOI table in Serosurveillance Base in Airtable (name, french translation and order).
+These records will be used for the "Population group" filters on serotracker.com.
+'''
+def ingest_sample_frame_goi_filter_options():
+    data = _airtable_get_request(AIRTABLE_SAMPLE_FRAME_GOI_OPTIONS_REQUEST_URL)
```
simonarocco, line 110: can we kill the naming convention of "_function name" for private functions? it was a pattern that I probably started but I don't think we use it consistently anymore so let's just keep everything formatted like "function_name" instead of "_function_name"

app/database_etl/etl_main.py
simonarocco: comment still has not been addressed i don't think @annoyingpr

app/database_etl/airtable_records_handler/airtable_records_getter.py
```diff
@@ -76,11 +100,39 @@ def get_all_records():
...
+                                                    name=record['fields']['Name'], 
+                                                    french_name=record['fields']['French Name'] if 'French Name' in record['fields'] else None, 
+                                                    order=record['fields']['Order'] if 'Order' in record['fields'] else None, 
+                                                    created_at=current_time) 
+                             for record in records if record['fields']]
+            session.bulk_save_objects(records_to_add)
+            session.commit()  
+        return 
+    except KeyError as e:
+       return _handle_airtable_error(e, data, AIRTABLE_SAMPLE_FRAME_GOI_OPTIONS_REQUEST_URL)
```
simonarocco, line 130: change from ```return _handle_airtable_error(e, data, AIRTABLE_SAMPLE_FRAME_GOI_OPTIONS_REQUEST_URL)``` to ```_handle_airtable_error(e, data, AIRTABLE_SAMPLE_FRAME_GOI_OPTIONS_REQUEST_URL)
return```. we don't want to return the return value of the function because downstream nothing is consuming it so we can just ```return``` which will return null

app/database_etl/airtable_records_handler/airtable_records_getter.py
```diff
@@ -19,6 +28,25 @@ def _add_fields_to_url(url):
...
+    headers = {'Authorization': 'Bearer {}'.format(AIRTABLE_API_KEY)}
+    body = "Results were not successfully retrieved from Airtable API. " \
+            "Please check connection parameters in config.py and fields in airtable_fields_config.json."
+    request_info = {
+        "url": url,
+        "headers": json.dumps(headers)
+    }
+    logging.error(e)
+    send_api_error_slack_notif(body, data, error=e, request_info=request_info, channel='#dev-logging-etl')
+    return request_info
```
simonarocco, line 48: if we aren't consuming ```request_info``` by the caller of this function we should just ```return```

simonarocco (commented):
see additional comments

app/database_etl/airtable_records_handler/airtable_records_getter.py
annoyingpr: ah oops, forgot to add the `AIRTABLE_API_KEY` for that one, just fixed it

app/database_etl/airtable_records_handler/airtable_records_getter.py
annoyingpr: cleaned it up for this file!

app/database_etl/etl_main.py
annoyingpr: `ingest_sample_frame_goi_options` doesn't return anything anymore - you can see in airtable_records_getter.py:127!

app/database_etl/airtable_records_handler/airtable_records_getter.py
annoyingpr: should i include a return after `handle_airtable_error(e, data, AIRTABLE_SAMPLE_FRAME_GOI_OPTIONS_REQUEST_URL, headers)` or should we be good?

app/database_etl/airtable_records_handler/airtable_records_getter.py
```diff
@@ -54,33 +80,58 @@ def _get_paginated_records(data, api_request_info):
...
+                                                    french_name=record['fields']['French Name'] if 'French Name' in record['fields'] else None, 
+                                                    order=record['fields']['Order'] if 'Order' in record['fields'] else None, 
+                                                    created_at=current_time) 
+                             for record in records if record['fields']]
+            session.bulk_save_objects(records_to_add)
+            session.commit()  
+        return 
+    except KeyError as e:
+       handle_airtable_error(e, data, AIRTABLE_SAMPLE_FRAME_GOI_OPTIONS_REQUEST_URL, headers)
+
```
simonarocco, line 130: add keyword return

simonarocco (approved):
small comment to address then can merge

PR #241, opened 2021-08-12
Simona/airtable test adj endpoint

simonarocco:
- Created an endpoint to run test adjustment given input parameters

https://airtable.com/tbli2lWQHAqBa6ZcI/viwDpHWCgFagItSc6/recKkt8WA13NFl1Cw?blocks=hide


app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -0,0 +1,262 @@
...
+    def __init__(self, model_code=testadj_model_code, model_name='testadj_binomial_se_sp',
+                 execution_params={}):
+        self.TESTADJ_MODEL = self.get_stan_model_cache(model_code=model_code, model_name=model_name)
+        self.TESTADJ_MODEL_NAME = model_name
+        # Parse execution params
+        self.n_iter = execution_params.get('n_iter', 2000)
+        self.n_chains = execution_params.get('n_chains', 4)
+        self.return_fit = execution_params.get('return_fit', False)
+        self.n_replicates = execution_params.get('n_replicates', 5)
+        self.trials_lim = execution_params.get('trials_lim', 100)
```
rahularoradfs, line 70: I think we decided that with Zayd's stability changes, we could bring this down to 5 trials and 1 model set.

@simonarocco, would this solve the AirTable automations issue?

rahularoradfs: @atatmaja

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -0,0 +1,262 @@
...
+            cached_model = pystan.StanModel(model_code=model_code, model_name=model_name, **kwargs)
+            with open(cache_fn, 'wb') as f:
+                pickle.dump(cached_model, f)
+        return cached_model
+
+    def fit_one_pystan_model(self, model_params: Dict) -> Tuple:
+        fit = self.TESTADJ_MODEL.sampling(data=model_params,
+                                          iter=self.n_iter,
+                                          chains=self.n_chains,
+                                          control={'adapt_delta': 0.95},
```
rahularoradfs, line 98: Shouldn't the hyperparameters here be different, @atatmaja @simonarocco? This is what Zayd was supposed to work on.

I think it was something like adapt_delta = 0.8, he set particular init values, etc... not sure that I see these reflected here

rahularoradfs (commented):
Just want to make sure that we include the hyperparameter tweaks that Zayd made in this PR. these were as follows, if I recall. See this link: https://public.3.basecamp.com/p/MjV8bWregyNEVduL9V4oW8Gy

- Adapt_delta = 0.99 if n_spec <10, n_sens <10, y_spec <5, or y_sens < 5
- n_chains = 3
- n_iter = 4000
- n_warmup = 1500
- run 5 trials, just with one model each - no reason to compare across modelsets
- don't check all the HMC diagnostics. Rather, just make sure that |Rhat| is bounded by 1.05; make sure that n_eff > 1000
- use the point estimates for sens, spec, prev as the inital values

app/namespaces/test_adjustment/test_adjustment_service.py
simonarocco: have we tested these changes actually result in this improvement? i can try it on airtable but we need to be certain that in no case will it exceed 30 seconds

app/namespaces/test_adjustment/test_adjustment_service.py
simonarocco: @atatmaja any idea on this? not sure at all, never saw his work

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -0,0 +1,262 @@
...
+
+            # Independent evaluation is available
+            if pd.notna(ind_se) and pd.notna(ind_sp):
+                adj_type = 'FINDDx / MUHC independent evaluation'
+                # Also note these must be divided by 100
+                se = ind_se
+                sp = ind_sp
+                output_se_n = float(ind_se_n) * 100 if ind_se_n is not None else 30
+                output_sp_n = float(ind_sp_n) * 100 if ind_sp_n is not None else 80
+
```
harrietware, line 194: Note that the research team is working on adding other independent evaluation sources to Airtable and updating how we want to prioritize independent, author, and manufacturer evaluation. Will let you know when this is in place

app/namespaces/test_adjustment/test_adjustment_service.py
simonarocco: @rahularoradfs I am unclear how to address the bottom 4 bullet points - can you plz be more specific about where in the code these changes need to be made?

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -0,0 +1,267 @@
...
+class TestAdjHandler:
+    def __init__(self, model_code=testadj_model_code, model_name='testadj_binomial_se_sp',
+                 execution_params={}):
+        self.TESTADJ_MODEL = self.get_stan_model_cache(model_code=model_code, model_name=model_name)
+        self.TESTADJ_MODEL_NAME = model_name
+        # Parse execution params
+        self.n_iter = execution_params.get('n_iter', 4000)
+        self.n_chains = execution_params.get('n_chains', 3)
+        self.return_fit = execution_params.get('return_fit', False)
+        self.n_replicates = execution_params.get('n_replicates', 5)
```
rahularoradfs, line 69: Okay, so. In the first instance, the algorithm that we were using was as follows:

- We were running the model 5x, and comparing the results between those 5 runs to see if it was consistent
- If it was consistent, we were taking the median result
- If it was not consistent, we were throwing it away and trying again - up to a total of 100 trials total, or 500 total model runs

This is a huge driver of why the test adjustment code was so slow. 

If we integrate some of the suggested optimizations, we should get a major, major speedup, that means that we can change our algorithm to the following:

- Run one model
- See whether it returns a sane result, with `result_is_bounded` and per the HMC diagnostics 
- If not, try running it again, up to a maximum of five models

I have highlighted the places where we can drop or simplify code in order to achieve this line by line below 
As a first step relevant to lines 69 and 70, we now no longer need the n_replicates param, and the default for trials_lim should be 5

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -0,0 +1,267 @@
...
+                pickle.dump(cached_model, f)
+        return cached_model
+
+    def fit_one_pystan_model(self, model_params: Dict) -> Tuple:
+        if model_params['n_sp' < 10] or model_params['n_se' < 10]  \
+                or model_params['y_sp' < 5]  or model_params['y_se' < 5]:
+            adapt_delta = 0.99
+        else:
+            adapt_delta = 0.8
+        fit = self.TESTADJ_MODEL.sampling(data=model_params,
```
rahularoradfs, line 100: This is where we can pass in the initial values. The documentation for how to pass in initial values is provided here: search for init. https://mc-stan.org/cmdstanr/reference/model-method-sample.html

A good starting format for the argument to pass in to init is as follows:

```
init_one_chain = {
    'prev': y_prev_obs / n_prev_obs,
    'sens': y_se / n_se,
    'spec': y_sp / n_sp
} # frequentist point estimates for each parameter as initial values 
init = [init_one_chain] * n_chains # pystan requires initial values to be expressed in a list of dicts, one dict for each chain
```

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -0,0 +1,267 @@
...
+        credible_interval_size = execution_params.get('credible_interval_size', 0.95)
+
+        # Validate model_params using marshmallow
+        try:
+            ModelParamsSchema().load(model_params)
+        except ValidationError as err:
+            print("Error: ", err.messages)
+            return None, None, None
+
+        satisfactory_model_set_found = False
```
rahularoradfs, line 134: We are going to running one model and seeing whether that single model is satisfactory, as opposed to running many model sets

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -0,0 +1,267 @@
...
+        n_modelsets = 0
+
+        while not satisfactory_model_set_found:
+            satisfactory_model_set = []
+            for _ in range(self.n_replicates):
```
rahularoradfs, line 139: In this case, we don't need to fit a model for each of n replicates - we just need to fit a single model

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -0,0 +1,267 @@
...
+                    if n_trials >= self.trials_lim:
+                        print('no models met the HMC diagnostics in {trials_lim} trials')
+                        return None, None, None
+
+                satisfactory_model_set.append(summary_df_parsed)
+
+            satisfactory_models_df = pd.DataFrame(satisfactory_model_set)
+
+            # see whether all results are within 10% of median; if not, rerun
+            model_medians = satisfactory_models_df['50%']
```
rahularoradfs, line 157: Then, we can get rid of the code from lines 154-160 that compare between model runs and get the median model run. We can just take the results from a single model

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -0,0 +1,267 @@
...
+            median = model_medians.median()
+            consistent_results_found = model_medians.between(0.9 * median, 1.1 * median).all()
+            median_run = satisfactory_models_df[satisfactory_models_df['50%'] == median].iloc[0]
+
+            lower, upper = arviz.hdi(median_run['samples'], credible_interval_size)
```
rahularoradfs, line 162: Need to make sure to keep lines 162-169 - we still need to calculate the confidence interval and check whether the result is bounded

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -0,0 +1,267 @@
...
+            best_fit = median_run['fit']
+
+            try:
+                raw_prev = model_params['y_prev_obs'] / model_params['n_prev_obs']
+                bounded = result_is_bounded(median, raw_prev)
+            except (ZeroDivisionError, ValueError):
+                return None, None, None
+
+            satisfactory_model_set_found = consistent_results_found and bounded
```
rahularoradfs, line 171: After we make the required changes, the biggest question will be whether the results from the model are bounded

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -0,0 +1,267 @@
...
+                                  index=summary['summary_rownames'],
+                                  columns=summary['summary_colnames'])
+
+        samples = fit.extract(pars='prev', permuted=True)['prev']
+
+        summary_df_parsed = summary_df.loc['prev', ['50%', '2.5%', '97.5%']]
+        summary_df_parsed['samples'] = samples
+        summary_df_parsed['fit'] = fit
+
+        diagnostics = pystan.diagnostics.check_hmc_diagnostics(fit)
```
rahularoradfs, line 117: This is what I was talking about when I said we don't need to check all of the diagnostics' values. We run the HMC diagnostics in line 117, which is great, but then we check to see whether all of them are true on line 119. We'd like to modify this such that we only check whether a couple of these are true.

My guess is that diagnostics is a dict, with keys that include something like 'rhat' and 'n_eff'.

- It might give you boolean values for both rhat and n_eff. If so, make sure that they're both true.
- It might give you numeric values for both. If so, make sure that `0.9 < Rhat < 1.1` and that `n_eff > 500`

rahularoradfs (commented):
@simonarocco left targeted comments on where to make the remaining changes! let me know if unclear - glad to hop on a call

app/namespaces/test_adjustment/test_adjustment_service.py
simonarocco: I don't get this comment - you want to pass in an additional parameter? What do you mean by pass in initial values? Right now we pass in data, iter, chains, control check_hmc_diagnostics.

rahularoradfs:
yes, an add'l param!

On August 19, 2021, simonarocco ***@***.***> wrote:
> @simonarocco commented on this pull request.
>

> In app/namespaces/test_adjustment/test_adjustment_service.py
> <https://github.com/serotracker/iit-
> backend/pull/241#discussion_r692592366>:
>

> + print(f"Using cached StanModel at filepath {cache_fn}") + + #
Otherwise create the model and cache it + except FileNotFoundError: +
cached_model = pystan.StanModel(model_code=model_code,
model_name=model_name, **kwargs) + with open(cache_fn, 'wb') as f: +
pickle.dump(cached_model, f) + return cached_model + + def
fit_one_pystan_model(self, model_params: Dict) -> Tuple: + if
model_params['n_sp' < 10] or model_params['n_se' < 10] \ + or
model_params['y_sp' < 5] or model_params['y_se' < 5]: + adapt_delta =
0.99 + else: + adapt_delta = 0.8 + fit =
self.TESTADJ_MODEL.sampling(data=model_params,
> I don't get this comment - you want to pass in an additional
> parameter? What do you mean by pass in initial values? Right now we
> pass in data, iter, chains, control check_hmc_diagnostics.
>

—
You are receiving this because you were mentioned.
<https://github.com/serotracker/iit-
backend/pull/241#discussion_r692592366>, or unsubscribe
auth/AOZY6J5KGPF4HUFBIEQOEU3T5WVFNANCNFSM5B76JMCQ>.
Triage notifications on the go with GitHub Mobile for iOS
<https://apps.apple.com/app/apple-store/id1477376905?ct=notification-
email&mt=8&pt=524675> or Android
<https://play.google.com/store/apps/details?id=com.github.android&utm_campaign=notification-
email>.

app/namespaces/test_adjustment/test_adjustment_controller.py
```diff
@@ -0,0 +1,66 @@
...
+from .test_adjustment_service import TestAdjHandler
+
+
+test_adjustment_ns = Namespace('test_adjustment', description='Endpoints for updating airtable records via scripting.')
+logging.getLogger(__name__)
+
+
+@test_adjustment_ns.route('/get_adjusted', methods=['POST'])
+class Records(Resource):
+    @test_adjustment_ns.doc('An endpoint for getting all records from database with or without filters.')
```
rahularoradfs, line 17: Is this the right documentation for this endpoint? Isn't this a test adj endpoint?

app/namespaces/test_adjustment/test_adjustment_controller.py
```diff
@@ -0,0 +1,66 @@
...
+            args=dict(request.args),
+            payload=data))
+
+        # Validate input payload
+        payload, status_code = validate_request_input_against_schema(data, TestAdjustmentSchema())
+        if status_code != 200:
+            # If there was an error with the input payload, return the error and 422 response
+            return make_response(payload, status_code)
+
+        # All of these params can be empty, in which case, our utility functions will just return all records
```
rahularoradfs, line 37: Why return all records? Shouldn't it just return the record of interest?

When we chat, let's chat about what this endpoint is for, exactly. What is the payload - shouldn't it just be a dict of the inputs required for test adjustment - and the result - shouldn't it just be posterior prev, sens, and spec?

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -0,0 +1,266 @@
...
+    y_sp = fields.Float()
+
+
+class TestAdjHandler:
+    def __init__(self, model_code=testadj_model_code, model_name='testadj_binomial_se_sp',
+                 execution_params={}):
+        self.TESTADJ_MODEL = self.get_stan_model_cache(model_code=model_code, model_name=model_name)
+        self.TESTADJ_MODEL_NAME = model_name
+        # Parse execution params
+        self.n_iter = execution_params.get('n_iter', 2000)
```
rahularoradfs, line 66: Might be able to drop this to 1e3 if this is still running slow. Let's leave it as is for now

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -0,0 +1,266 @@
...
+        self.n_chains = execution_params.get('n_chains', 4)
+        self.return_fit = execution_params.get('return_fit', False)
+        self.trials_lim = execution_params.get('trials_lim', 5)
+        self.modelsets_lim = execution_params.get('modelsets_lim', 3)
```
rahularoradfs, line 70: Didn't we decide to get rid of the modelsets side of this altogether? Get rid of that outer loop?

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -0,0 +1,266 @@
...
+        try:
+            ModelParamsSchema().load(model_params)
+        except ValidationError as err:
+            print("Error: ", err.messages)
+            return None, None, None
+
+        satisfactory_model_set_found = False
+        n_modelsets = 0
+
+        while not satisfactory_model_set_found:
```
rahularoradfs, line 147: We can probably get rid of the model_set logic altogether

rahularoradfs (approved):

PR #242, opened 2021-08-13
Remove duplicate household and community samples from static filter options

simonarocco:
- "Household and community samples" was repeated twice in the static filter options for population group
-


brett-stephen (approved):
looks good, Thanks Simona!

PR #243, opened 2021-08-16
Changing Gsheet column range from 10k to 100k

simonarocco:
- Changing Gsheet column range from 10k to 100k

- Ran ETL, didn't get any errors, saw that data was properly loaded in sheet


rahularoradfs (approved):
thanks for handling!

PR #250, opened 2021-09-10
Update country mapping

harrietware:
Update country mapping to WHO standards (see email chain forwarded).

N/A

N/A

N/A


simonarocco (approved):
don't need the "Code" column because we already store the iso3 codes in a separate csv! otherwise looks good

PR #251, opened 2021-09-11
Add new test adj model to ETL

simonarocco:
- Some updates to the test adj model and sanity checks to make sure certain params are bounded correctly
- Making ETL use new test adj model that is being used by endpoint
- Moved the function in the ETL that applies diffing logic and runs test adj to the test adj folder in ETL (makes more sense to have it here)

https://airtable.com/appT8tnLbGJV6v81V/tbli2lWQHAqBa6ZcI/viwDpHWCgFagItSc6/recyBq9zzIAb2kAVb?blocks=hide

- Ran the ETL locally for 48 estimates, sanity checked outputs

- Once this PR is approved, will need to run ETL on ALL estimates in prod

### DIFF
--- a/app/database_etl/test_adjustment_handler/test_adjustment.py
+++ b/app/database_etl/test_adjustment_handler/test_adjustment.py
@@ -108,14 +108,6 @@ def add_test_adjustments(df: pd.DataFrame) -> pd.DataFrame:
     # Get all unique airtable_record_ids that are new/have been modified
     new_airtable_record_ids = diff['airtable_record_id'].unique()
 
-    # Add all unique airtable_record_ids for which test adjustment was unsuccessful
-    # TODO: MUST BE COMMENTED OUT IN PROD
-    """
-    unadjusted_airtable_record_ids = total_db_records[total_db_records['adj_prevalence'].isna()]['airtable_record_id'].unique()
-    new_airtable_record_ids = set.union(set(new_airtable_record_ids), 
-                                        set(unadjusted_airtable_record_ids))
-    """
-
     # Get all rows from airtable data that need to be test adjusted, and ones that don't
     old_airtable_test_adj_records = \
         df[~df['airtable_record_id'].isin(new_airtable_record_ids)].reset_index(
--- a/app/database_etl/test_adjustment_handler/test_adjustment.py
+++ b/app/database_etl/test_adjustment_handler/test_adjustment.py
@@ -1,20 +1,16 @@
 # monte-carlo version of RG estimator,
 # accounting for uncertainty in se and sp estimates
-import pickle
 from math import log
-from hashlib import md5
-from typing import Dict, Tuple, Union
 
 import pandas as pd
-import pystan
-import arviz
-from statsmodels.stats.proportion import proportion_confint
 from marshmallow import Schema, fields, ValidationError
 
-from app.database_etl.test_adjustment_handler import bastos_estimates, testadj_model_code
+from app.database_etl.test_adjustment_handler import testadj_model_code
+from app.namespaces.test_adjustment import TestAdjHandler
+from app.serotracker_sqlalchemy import db_session, ResearchSource, DashboardSource
 
 
-def logit(p, tol = 1e-3):
+def logit(p, tol=1e-3):
     # logit function; well defined for p in open interval (0,1)
     
     # constrain the probability to the range [tol, 1-tol]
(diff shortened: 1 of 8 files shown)


app/namespaces/test_adjustment/test_adjustment_schema.py
```diff
@@ -3,12 +3,14 @@
 
 class TestAdjustmentSchema(Schema):
     test_adj = fields.Boolean(allow_none=True)
+    # ind_se and ind_sp are percentages; and must have values between 0.005
```
rahularoradfs, line 6: between 0.005?

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -233,19 +240,32 @@ def get_adjusted_estimate(self, test_adj, ind_se, ind_sp, ind_se_n, ind_sp_n,
...
-                    print(f'FAILED TO ADJUST ESTIMATE')
-
+                # Check bounds of se, se_n, sp, sp_n
+                if se and not (se <= 1 and se >= 0.005):
+                    logging.error("se is not between 0.005 and 1")
+                    lower, adj_prev, upper = None, None, None
+                elif sp and not (sp <= 1 and sp >= 0.005):
+                    logging.error("sp is not between 0.005 and 1")
+                    lower, adj_prev, upper = None, None, None
+                elif output_se_n and output_se_n <= 1:
```
rahularoradfs, line 250: These don't need to be bounded between 1 and 1000 - just > 1 is fine

rahularoradfs (commented):
Overall checks out to me! Might be worth it to get a more careful review from a BE dev.

app/namespaces/test_adjustment/test_adjustment_service.py
simonarocco: what?? it is not bounded between 1 and 1000. im just making sure that output_se_n is not null and then checking if it's greater than 1

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -233,19 +240,32 @@ def get_adjusted_estimate(self, test_adj, ind_se, ind_sp, ind_se_n, ind_sp_n,
...
+                    logging.error("se_n is not between 1 and 100")
+                    lower, adj_prev, upper = None, None, None
+                elif output_sp_n and output_sp_n <= 1:
+                    logging.error("sp_n is not between 1 and 100")
```
atatmaja, line 254: The log here implies that sp_n should be between 1 and 100, although we seem to only care about it being less than 1. Can we update the logs to be consistent with what we actually need to validate

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -73,19 +74,26 @@ def __init__(self, model_code=testadj_model_code, model_name='testadj_binomial_s
...
+        abs_filepath_curr_dir = os.getcwd()
+        proj_root_abs_path = abs_filepath_curr_dir.split("iit-backend")[0]
+
+        # Create model absolute path
         code_hash = md5(model_code.encode('ascii')).hexdigest()
-        cache_fn = f'app/namespaces/test_adjustment/stanmodelcache-{model_name}-{code_hash}.pkl'
+        cache_fn = f'{proj_root_abs_path}iit-backend/app/namespaces/test_adjustment/stanmodelcache-{model_name}-{code_hash}.pkl'
+
+        # Get relative path of model cache (necessary because funtion will be called from different files)
+        rel_path = os.path.relpath(cache_fn, abs_filepath_curr_dir)
```
atatmaja, line 86: I don't get why having the relative path is advantageous over having the absolute path if we're trying to refer to this cache from different files. In fact, isn't having the relative path more confusing and hard to work with?

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -111,8 +119,7 @@ def fit_one_pystan_model(self, model_params: Dict) -> Tuple:
                                           iter=self.n_iter,
                                           chains=self.n_chains,
                                           control={'adapt_delta': adapt_delta},
-                                          check_hmc_diagnostics=False,
-                                          init=init)
```
atatmaja, line 115: why did we scrap this `init`?

app/database_etl/test_adjustment_handler/test_adjustment.py
```diff
@@ -270,3 +63,122 @@ def run_on_test_set(model_code: str = testadj_model_code, model_name: str = 'tes
...
+    # Drop duplicates based on these cols
+    duplicate_cols = ['airtable_record_id', 'test_adj', 'ind_se', 'ind_sp', 'ind_se_n', 'ind_sp_n',
+                      'se_n', 'sp_n', 'sensitivity', 'specificity', 'test_validation', 'test_type', 'denominator_value',
+                      'serum_pos_prevalence']
+    diff = diff.drop_duplicates(subset=duplicate_cols, keep=False)
+
+    # Get all unique airtable_record_ids that are new/have been modified
+    new_airtable_record_ids = diff['airtable_record_id'].unique()
+
+    # Add all unique airtable_record_ids for which test adjustment was unsuccessful
```
atatmaja, line 111: Are we ever using this commented out code anymore? If not can we get rid, it's an eyesore.

If do need this logic to rerun test adjustment for data refresh purposes, I would personally prefer keeping it out of this function (maybe make a separate script/utility to do this)

atatmaja (commented):
Few questions & nits

app/namespaces/test_adjustment/test_adjustment_service.py
simonarocco: because it was breaking the initialization of test adj in some cases and we dont know why and rahul said it was fine to leave out

app/database_etl/test_adjustment_handler/test_adjustment.py
simonarocco: took it out. we won't be doing this in the etl

app/namespaces/test_adjustment/test_adjustment_service.py
simonarocco: changed to just use abs path.

app/namespaces/test_adjustment/test_adjustment_service.py
simonarocco: done

atatmaja (approved):
LGTM, just make sure it's tested

PR #252, opened 2021-09-21
Github Data Repository - Create Python Driver Script

annoyingpr:
[Github Data Repository - Create Python Driver Script](https://airtable.com/appT8tnLbGJV6v81V/tbli2lWQHAqBa6ZcI/viwTYqtBDdt3Wt3Yo/recbsNmzqAsqbejva?blocks=hide)

- Run `python3 csv_download_script.py`

### DIFF
--- a/csv_download_script.py
+++ b/csv_download_script.py
@@ -1,0 +1,49 @@
+import pandas as pd 
+import requests
+import collections
+import os
+
+AIRTABLE_API_KEY = os.getenv('AIRTABLE_API_KEY')
+
+headers = {'Authorization': 'Bearer {}'.format(AIRTABLE_API_KEY)}
+url = "https://api.airtable.com/v0/appFfYIupTrVQ0HJx/Rapid%20Review%3A%20Estimates?view=SeroTracker:%20Serosurveys%20Reporting%20Prevalence"
+
+
+wanted_fields = ['Prevalence%20Estimate%20Name', 'Publication%20Date', "Grade%20of%20Estimate%20Scope", "Country", 
+                    "Specific%20Geography", "Sampling%20Start%20Date", "Sampling%20End%20Date", "Sample%20Frame%20(groups%20of%20interest)", 
+                    "Sample%20Frame%20(age)", "Denominator%20Value", "Serum%20positive%20prevalence", "Serum%20pos%20prevalence,%2095pct%20CI%20Upper", 
+                    "Serum%20pos%20prevalence, 95pct%20CI%20Lower", "Sampling%20Method", "Test%20Manufacturer", "Test%20Type", "Isotype(s)%20Reported",
+                    "Sensitivity%20(from%20Test%20Evaluation%20lookup)", "Specificity%20(from%20Test%20Evaluation%20lookup)", "Overall%20Risk%20of%20Bias%20(JBI)",
+                    "Source%20Type", "First%20Author%20Full Name", "Lead%20Institution", "URL", "Date%20Created", "Last%20modified%20time", "Data%20Quality%20Status"]
+
+# Adding all wanted fields as query params to URL
+for wanted_field in wanted_fields:
+    url += "&fields=" + wanted_field
+
+organized_data = collections.defaultdict(list)
+offset = ""
+
+print("Starting CSV download...")
+
+while True:
+    # Add offset as query param to URL if there is a valid offset
+    new_url = url + "&offset=" + offset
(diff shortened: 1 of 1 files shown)


csv_download_script.py
```diff
@@ -0,0 +1,49 @@
...
+# Adding all wanted fields as query params to URL
+for wanted_field in wanted_fields:
+    url += "&fields=" + wanted_field
+
+organized_data = collections.defaultdict(list)
+offset = ""
+
+print("Starting CSV download...")
+
+while True:
```
atatmaja, line 28: Doesn't `get_paginated_records` already do lots of this functionality (https://docs.github.com/en/actions/reference/events-that-trigger-workflows#schedule)?? Why didn't we just leverage the functions we already had?

csv_download_script.py
```diff
@@ -0,0 +1,49 @@
...
+    if "offset" in data:
+        offset = data['offset']
+    else:
+        break
+    
+# Put into dataframe
+df = pd.DataFrame.from_dict(organized_data, orient="index")
+
+# Output to a csv file 
+df.to_csv("serotracker_data_download.csv")
```
atatmaja, line 46: we probably don't want an index column to be in our public facing CSV right? in that case I think we could supply the argument `index=false` here (https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_csv.html)

atatmaja:
I think we should define this script in some folder as we will probably have other scripts related to our data download operations, e.g. stuff pertaining to changelogs. We could call it `data_download_manager` or `github_upload_manager` or something, idk I'm not very good at names lol.

csv_download_script.py
simonarocco: agreed! as we discussed prannoy it would be great if we can just leverage the existing functions we already have that we use in the etl

csv_download_script.py
```diff
@@ -0,0 +1,49 @@
...
+import collections
+import os
+
+AIRTABLE_API_KEY = os.getenv('AIRTABLE_API_KEY')
+
+headers = {'Authorization': 'Bearer {}'.format(AIRTABLE_API_KEY)}
+url = "https://api.airtable.com/v0/appFfYIupTrVQ0HJx/Rapid%20Review%3A%20Estimates?view=SeroTracker:%20Serosurveys%20Reporting%20Prevalence"
+
+
+wanted_fields = ['Prevalence%20Estimate%20Name', 'Publication%20Date', "Grade%20of%20Estimate%20Scope", "Country",
```
simonarocco, line 12: Look at how we append the fields to the airtable get request URL in the existing functionality. Looks a bit odd saving a list with all of these strings. We probably want to do what we already have which is save a list with the normal field names, then when we loop through and add to the URL we put in HTML format. Take a look at the functions here: https://github.com/serotracker/iit-backend/blob/master/app/database_etl/airtable_records_handler/airtable_records_getter.py

csv_download_script.py
```diff
@@ -0,0 +1,49 @@
...
+                    "Source%20Type", "First%20Author%20Full Name", "Lead%20Institution", "URL", "Date%20Created", "Last%20modified%20time", "Data%20Quality%20Status"]
+
+# Adding all wanted fields as query params to URL
+for wanted_field in wanted_fields:
+    url += "&fields=" + wanted_field
+
+organized_data = collections.defaultdict(list)
+offset = ""
+
+print("Starting CSV download...")
```
simonarocco, line 26: Probably don't need this. Would be more effective to think about error logging for this entire procedure. Where do we want to pipe errors to? How will the dev team know if the upload failed? Let's get thinking about this and then we can integrate error handling/logging into the script

simonarocco (commented):
Great start! Overall main comments:

1) Leverage existing functionality!! This script can be really small but just using existing functions we already have.

2) Where is this script saved? Recall our discussion about wanting to create a repo for the entire GitHub CSV project - this script should be saved in that repo, along with README and changelog as @atatmaja also mentioned. Let's create that repo and have it be **separate** from iit-backend. Remember, this is going to be a public facing repo, it shouldn't be within the rest of our backend. Separating this out into a separate repo will have implications for using functions/utils in iit-backend, but there are ways around that. Take a look at GitHub submodules for this, which we have previously used.

3) We need to be thoughtful about error handling and logging. When this script fails, will it fail safely? How will it fail? Who will be notified to fix it and how? These are all considerations that we should have in the full tech design document for this project.

simonarocco:
Closing pull request as I am now taking over this PR

PR #253, opened 2021-09-23
Creating python script to generate github CSV

simonarocco:
- This PR creates the python script that will be run by the GitHub action to create the CSV we will upload to our public repo
- The script 1) Reads which fields the CSV should contain from the table in Airtable called "GitHub CSV Fields" and then 2) Reads from the Estimates table in Airtable to pull records for the specified fields

A few things to note in this PR:

1) I have modified the very important get_all_records() function to now include a fields parameter so that the ETL and this driver script can both use this function by just passing in diff fields to use when scraping the estimates table. I ran ETL with this new change to confirm it works and we can still pull records properly.

2) I added three new fields to the airtable_fields_config as these are additional fields we want to show in our CSV but that we previously didn't have in this config. We use this config because in the CSV we want to format column names in a dev friendly way (owid does the same).

https://airtable.com/appT8tnLbGJV6v81V/tbli2lWQHAqBa6ZcI/viwDpHWCgFagItSc6/recS2F1JxIB3ZFjJg?blocks=hide

- Ran the script and made sure CSV was properly generated and looked good
- Also ran ETL to make sure no impacts on that

- Yes need to set up github actions to run this script which includes the appropriate secrets and env variables


app/database_etl/airtable_records_handler/airtable_records_getter.py
```diff
@@ -77,11 +76,11 @@ def get_paginated_records(data, api_request_info):
...
 
 
-def get_all_records():
+def get_all_records(fields):
     # Get airtable API URL and add fields to be scraped to URL in HTML format
     headers = {'Authorization': 'Bearer {}'.format(AIRTABLE_API_KEY)}
     url = AIRTABLE_REQUEST_URL.format(AIRTABLE_BASE_ID)
-    url = add_fields_to_url(url)
+    url = add_fields_to_url(url, fields)
     url += '&filterByFormula={ETL Included}=1'
```
atatmaja, line 84: Will this `filterByFormula` parameter be needed when we use this function to request for the github CSV?

atatmaja: Maybe we could make this function also take in an optional "filter" parameter, which would allow you to append any filter strings onto the url. 

This would make this function super generalizable for all airtable requests (including our normal airtable records request, github csv request, and filter options request)

app/github_public_repo/estimate_csv_creator.py
```diff
@@ -0,0 +1,32 @@
...
+
+# Get csv columns and create csv
+try:
+    # Extract fields
+    records = data['records']
+    fields = [x['fields']['Field Name'] for x in records]
+    logging.info("Successfully retrieved field names from Airtable")
+
+    # Get estimates from Rapid Review: Estimates table for specified fields
+    csv_records = get_all_records(fields)
```
atatmaja, line 24: I think get_all_records already implements error logging. With that in mind, we should move the try except to after this line IMO so we can log a more targeted error message

atatmaja (commented):
Some nits but overall fine

app/github_public_repo/estimate_csv_creator.py
```diff
@@ -0,0 +1,32 @@
...
+
+    # Convert to df, reorder cols alphabetically and as save as csv
+    csv_records_df = pd.DataFrame.from_dict(csv_records)
+    csv_records_df = csv_records_df.reindex(sorted(csv_records_df.columns), axis=1)
+    csv_records_df.to_csv('GitHub CSV.csv', index=False)
```
atatmaja, line 29: Will there be any issues with specifying the output file path like this (i.e. if we invoke this script from a few different python paths)

app/database_etl/airtable_records_handler/airtable_records_getter.py
simonarocco: I thought about that and I'm pretty sure we do want this filter by formula to be included because it means it will only show records that haven't been superceded, ones where the denom isn't null, and ones where serum pos prev isn't null which is what we want to show i think. I do like the idea of adding a filter param though so I will go ahead and add that.

app/github_public_repo/estimate_csv_creator.py
simonarocco: i don't think so, it will just save the csv under this name in the same directory wherever the script is being called right. this means that the script caller just needs to keep track of where it will be saved. where will the script be called from when we run it as a github actions @annoyingpr ? we will need to run the script, generate the csv, and then know the file path of the script so we can upload the csv to the repo.

app/database_etl/airtable_records_handler/airtable_records_getter.py
simonarocco: Hm, I'm actually going to go back on this. We don't use this function for get filter records - it's currently only being used by our normal ETL and by the new github csv script. So right now as it stands, no other pieces of the code use this, and these two applications require the etl included filter. So I can make it a default filters = &filterByFormula={ETL Included}=1' param for now, and then in the future if we need this function it is already generalized. does that make sense?

app/github_public_repo/estimate_csv_creator.py
simonarocco: It does implement error logging, but all it does is it logs the KeyError e. So if we have get all records inside this try except, and it encounters the keyerror inside this function, it will pass the exact same error back to the except and the same 'e' will be logged just as it would be inside the get all records function. Does that make sense? Basically the keyerror will bubble up back to the original try except and it will still be logged with the same granularity. See handle_airtable_error on line 36 of airtable_records_getter

app/github_public_repo/README.md
```diff
@@ -0,0 +1,20 @@
+
+# SARS-CoV-2 Serosurveillance Data by Serotracker
+
+SeroTracker synthesizes findings from thousands of COVID-19 seroprevalence studies worldwide, providing a data platform and interactive dashboard for SARS-CoV-2 serosurveillance. This dataset represents our collection of serosurveillance studies
```
simonarocco, line 4: Let's add in the link to our website (maybe hyperlink interactive dashboard). Also change:

"The dataset in <insert name of csv for them to find> represents our collection of serosurveillance studies."

app/github_public_repo/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Download
+Our complete SARS-Cov-2 dataset is available as a CSV programmatically with the following command:
```
simonarocco, line 7: Change:

Our complete SARS-Cov-2 dataset can be scraped programmatically from this repository in the following ways:

1) Using the command: `curl https://raw.githubusercontent.com/serotracker/sars-cov-2-data/main/test.csv`
2) Using a GET request through REST API protocol

app/github_public_repo/README.md
```diff
@@ -0,0 +1,20 @@
...
+- `curl https://raw.githubusercontent.com/serotracker/sars-cov-2-data/main/test.csv`.
+
+Otherwise, you can also download our dataset directly from our [Airtable view]([https://airtable.com/shraXWPJ9Yu7ybowM/tbljN2mhRVfSlZv2d?backgroundColor=blue&viewControls=on](https://airtable.com/shraXWPJ9Yu7ybowM/tbljN2mhRVfSlZv2d?backgroundColor=blue&viewControls=on)).
```
simonarocco, line 10: Change:

Our dataset can also be downloaded directly from our...

app/github_public_repo/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+The CSV file follows the format of 1 row per serosurveillance study.
+
+  
+### Additional Information
+Please see our [Data Dictionary]([https://docs.google.com/spreadsheets/d/1KQbp5T9Cq_HnNpmBTWY1iKs6Etu1-qJcnhdJ5eyw7N8/edit#gid=0](https://docs.google.com/spreadsheets/d/1KQbp5T9Cq_HnNpmBTWY1iKs6Etu1-qJcnhdJ5eyw7N8/edit#gid=0)) for explanations of our variables, data types, and descriptions as well as insight into how our data is collected by our research team.
```
simonarocco, line 16: @annoyingpr please confirm that our data dictionary has all the columns that were are showing in the csv.

app/github_public_repo/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+In order to keep up to date with important changes to our dataset, please consult our [Change Log]([https://airtable.com/shrxpAlF6v0LeRYkA/tblC6jj904WXUzwVY](https://airtable.com/shrxpAlF6v0LeRYkA/tblC6jj904WXUzwVY)) regularly.
```
simonarocco, line 18: You copied the URL twice here

simonarocco: Are we sure that this change log will capture all of the changes relevant to the CSV? i.e. there isn't a case in which there will be a change made to the csv that won't be represented in the change log? Just want to make sure we have confirmed this with the research team.

simonarocco (commented):
readme looks good, just a few questions/changes requested. is the github action just purely set up through github or do we need to create any kind of file for it?

annoyingpr:
> readme looks good, just a few questions/changes requested. is the github action just purely set up through github or do we need to create any kind of file for it?

Github action is setup in the `.github/workflows` folder and it's called `push-csv-to-repo.yml`!

app/github_public_repo/README.md
annoyingpr: seems like we're missing two columns: `Last Modified Time` and `Data Quality Status`. i've messaged Christian about including them!

app/github_public_repo/README.md
annoyingpr: yup! confirmed with Christian who will be one of the main maintainers

app/github_public_repo/estimate_csv_creator.py
atatmaja: okay just checking, will leave it up to @annoyingpr to ensure this won't be an issue

app/github_public_repo/estimate_csv_creator.py
atatmaja: sure, it'll log twice in that scenario (a very very nitpicky concern tho, in fact probably not even worth)

app/github_public_repo/README.md
```diff
@@ -0,0 +1,20 @@
...
+Our dataset can also be downloaded directly from our [Airtable view]([https://airtable.com/shraXWPJ9Yu7ybowM/tbljN2mhRVfSlZv2d?backgroundColor=blue&viewControls=on](https://airtable.com/shraXWPJ9Yu7ybowM/tbljN2mhRVfSlZv2d?backgroundColor=blue&viewControls=on)).
+
+The CSV file follows the format of 1 row per serosurveillance study.
+ 
+### Additional Information
+Please see our [Data Dictionary]([https://docs.google.com/spreadsheets/d/1KQbp5T9Cq_HnNpmBTWY1iKs6Etu1-qJcnhdJ5eyw7N8/edit#gid=0](https://docs.google.com/spreadsheets/d/1KQbp5T9Cq_HnNpmBTWY1iKs6Etu1-qJcnhdJ5eyw7N8/edit#gid=0)) for explanations of our variables, data types, and descriptions as well as insight into how our data is collected by our research team.
+
+In order to keep up to date with important changes to our dataset, please consult our [Change Log](https://airtable.com/shrxpAlF6v0LeRYkA/tblC6jj904WXUzwVY) regularly.
+
+For more information about how we collect, extract and use our data, please see the [Data page]([https://serotracker.com/en/Data](https://serotracker.com/en/Data)) on our website.
```
atatmaja, line 20: Some suggestions of things to add (some based on OWIDs readme):

- Information on how to cite us, licensing if necessary, etc
- Form for submitting sources
- Form for people who want to contact us (similar to data download, will give us a chance to collect contact info about people who are down to use our data)

@simonarocco @annoyingpr 

cc @rahularoradfs

atatmaja (approved):

app/github_public_repo/estimate_csv_creator.py
annoyingpr: the script will most likely be running from root, but currently trying to figure out if we can setup an absolute path for this CSV

app/github_public_repo/README.md
annoyingpr: made a [new form](https://forms.gle/gqi3kvKQgasYCrQE9) for this, and pushed updates based on your suggestions!

jordanvanwyk:
Export CSV and check all columns are still being updated/in use.
Check that every column in the CSV has a pair in the data dictionary.
I added some additional fields to the CSV... eventually we need to have the ones that are in the database too (e.g., case, deaths, vaccines).
@simonarocco and @MaireadWhelan can likely help with these items.

app/github_public_repo/estimate_csv_creator.py
```diff
@@ -0,0 +1,34 @@
+import logging
+import os
+import pandas as pd
+from flask import current_app as app
+
+from app.database_etl.airtable_records_handler import get_all_records, airtable_get_request
+
+# Load env variables for Airtable request
+AIRTABLE_API_KEY = app.config['AIRTABLE_API_KEY']
+AIRTABLE_GITHUB_CSV_FIELDS_REQUEST_URL = app.config['AIRTABLE_GITHUB_CSV_FIELDS_REQUEST_URL']
```
rahularoradfs, line 10: This is great. Do we have a process for who, on the research side, is keeping this up to date? Is this @123chrisc or have we delegated the responsibility elsewhere?

app/github_public_repo/estimate_csv_creator.py
```diff
@@ -0,0 +1,34 @@
...
+try:
+    # Extract fields
+    records = data['records']
+    fields = [x['fields']['Field Name'] for x in records]
+    logging.info("Successfully retrieved field names from Airtable")
+
+    # Get estimates from Rapid Review: Estimates table for specified fields
+    csv_records = get_all_records(fields)
+
+    # Convert to df, reorder cols alphabetically and as save as csv
```
rahularoradfs, line 26: Can we do better than reordering cols alphabetically here? There are certain columns which index the data - country, sampling dates, etc. Seems like we ought to arrange them in this order.

We can arrange them in AirTable in this fashion, and then order them in this CSV in the same fashion.

I know it's programatically equivalent -- but oftentimes, when working with data like this, I'll manually look at the CSV to see how it's laid out. and that's much easier if it's laid out in an order that makes sense to me

app/github_public_repo/estimate_csv_creator.py
rahularoradfs: Also, have we double checked that this contains all the fields that we want to push for analysis?

app/github_public_repo/estimate_csv_creator.py
```diff
@@ -0,0 +1,34 @@
...
+    csv_records_df = pd.DataFrame.from_dict(csv_records)
```
rahularoradfs, line 27: Which rows are included here? All rows, or only primary estimates? If all rows, have we included the identifiers that distinguish between two highly similar estimates so folks can easily identify what the differences are? Have we provided the guidance in the README that indicates this structure, and why there are many estimates within each study?

app/github_public_repo/README.md
```diff
@@ -0,0 +1,26 @@
...
+[SeroTracker](https://serotracker.com/en/Explore) synthesizes findings from thousands of COVID-19 seroprevalence studies worldwide, providing a data platform and interactive dashboard for SARS-CoV-2 serosurveillance. This dataset in this repository represents our collection of serosurveillance studies.
+
+### Download
+Our complete SARS-Cov-2 dataset can be scraped programmatically from this repository in the following ways:
+1. Using the command: `curl https://raw.githubusercontent.com/serotracker/sars-cov-2-data/main/test.csv`.
+2. Using a GET request through REST API protocol.
+
+Our dataset can also be downloaded directly from our [Airtable view]([https://airtable.com/shraXWPJ9Yu7ybowM/tbljN2mhRVfSlZv2d?backgroundColor=blue&viewControls=on](https://airtable.com/shraXWPJ9Yu7ybowM/tbljN2mhRVfSlZv2d?backgroundColor=blue&viewControls=on)).
+
+The CSV file follows the format of 1 row per serosurveillance study.
```
rahularoradfs, line 13: Is this the case? Are we only pushing primary estimates from each study? Or are we pushing one row per estimate, and each row is one seroprevalence estimate - that is, are we making subgroup data available?

app/github_public_repo/README.md
rahularoradfs: Does the data dictionary have the `database_names` for each variable?

rahularoradfs (commented):
From a code and implementation perspective, this looks very good to me. The question coming to mind here is the user experience. Is this something that provides everything that a new analyst needs to start using our data from scratch? Does it integrate well with our data dictionary? To ensure this, there's another set of steps we need to take here, which involve getting detailed review on this from consumers of our data 

The two folks who I think should review are Harriet (as the primary consumer of our data) and Julia Ma (as a close collaborator, semi-external, who also has used our data pretty extensively). I would love to get their detailed comments on three things:
@harrietware, it would be great to get your take on a few things here, as a heavy user of data sources like this, and as the primary person analyzing our data internally right now. Three things come to mind here:
1. Which fields we need to include in the first instance. It'd probably be easiest for them to provide this in terms as a list of the fields, with their `database_names`, that they find most important for analysis
2. How we should order these fields to be most intuitive
3. What needs to be included in the README for this to be useful to each of them as analysts. What are best practices from other repos? Eg do we need a codebook? and so on


We can either push now and get that detailed review in the next iteration, or get that detailed review before merging. Let me know what you decide. I am happy to put you in touch with Harriet and Julia to request that level of detailed review

app/github_public_repo/estimate_csv_creator.py
annoyingpr: I will talk to @123chrisc about being the DRI or assigning this responsibility!

app/github_public_repo/README.md
annoyingpr: right now the names are the same, but are not reflected in snake case (e.g. right now it is "Prevalence Estimate Name" vs "prevalence_estimate_name")? @rahularoradfs should this be updated? i can pass on to @123chrisc who's doing the data dict updates

app/github_public_repo/estimate_csv_creator.py
simonarocco: this is resolved. we are now using abs path.

app/github_public_repo/README.md
simonarocco: As discussed in the basecamp thread in the Data Infrastructure group, we want our data dictionary to contain all the snake case names of the human readable names. For example, we want our data dict to include both "Prevalence Estimate Name" and "prevalence_estimate_name" so it is easy for users to map these. @annoyingpr can you:

1) make sure all the fields in this airtable table are in the data dictionary. If there are missing fields, please add them to the data dict https://airtable.com/appFfYIupTrVQ0HJx/tbliKC06AgJ1EAKLa/viwoxvB5i6kITkaiF?blocks=hide

2) make sure all the snake casing of fields is added to our data dictionary. you can use the 'full_airtable_fields' dict in this file to create this mapping correctly: https://github.com/serotracker/iit-backend/blob/master/app/utils/airtable_fields_config.py

app/github_public_repo/README.md
simonarocco: Updated: the readme changelog will now be maintained by gabe on the research team who will also be maintaining the corresponding airtable table containing the field names.

app/github_public_repo/estimate_csv_creator.py
simonarocco: Resolving this conversation as Gabe will now be the maintainer of this table.

app/github_public_repo/estimate_csv_creator.py
simonarocco: ya, will fix this. we will keep the column ordering in: https://airtable.com/appFfYIupTrVQ0HJx/tbliKC06AgJ1EAKLa/viwoxvB5i6kITkaiF?blocks=hide

app/github_public_repo/README.md
simonarocco: No, this file is wrong. One row corresponds to one estimate. We are showing all estimates not just primary estimates. Need to update this.

PR #268, opened 2021-09-27
Christian/update estimate prioritization

123chrisc:
This PR addresses (slightly) outdated estimate prioritization code
N/A?
I reviewed the estimate prioritization code and compared with our SOP and protocol prioritization. After making comparisons, I made minor edits to match.
No?

### DIFF
--- a/app/utils/estimate_prioritization/prioritization_criteria.py
+++ b/app/utils/estimate_prioritization/prioritization_criteria.py
@@ -7,8 +7,8 @@ prioritization_criteria_full = {
     ],
     'adjustment_testadj': [
         lambda estimate: (estimate['pop_adj'] is True) and (estimate['test_adj'] is True),
-        lambda estimate: (pd.isna(estimate['pop_adj'])) and (estimate['test_adj'] is True),
         lambda estimate: (estimate['pop_adj'] is True) and (pd.isna(estimate['test_adj'])),
+        lambda estimate: (pd.isna(estimate['pop_adj'])) and (estimate['test_adj'] is True),
         lambda estimate: (pd.isna(estimate['pop_adj'])) and (pd.isna(estimate['test_adj'])),
     ],
     'primary_estimate_testunadj': [
@@ -52,8 +52,11 @@ prioritization_criteria_full = {
     ],
     'test_type': [
         lambda estimate: estimate['test_type'] == 'Neutralization',
+        lambda estimate: estimate['test_type'] == 'Multiple Types',
+        lambda estimate: estimate['test_type'] == 'ELISA',
         lambda estimate: estimate['test_type'] == 'CLIA',
-        lambda estimate: estimate['test_type'] == 'ELISA'
+        lambda estimate: estimate['test_type'] == 'LFIA',
+        lambda estimate: estimate['test_type'] == 'Other',
     ],
     'specimen': [
         lambda estimate: estimate['specimen_type'] != 'Dried Blood'
--- a/app/utils/estimate_prioritization/prioritization_criteria.py
+++ b/app/utils/estimate_prioritization/prioritization_criteria.py
@@ -24,7 +24,6 @@ prioritization_criteria_full = {
         lambda estimate: estimate['estimate_grade'] == 'National',
         lambda estimate: estimate['estimate_grade'] == 'Regional',
         lambda estimate: estimate['estimate_grade'] == 'Local',
-        lambda estimate: estimate['estimate_grade'] == 'Sublocal',
     ],
     'age': [
         lambda estimate: estimate['age'] == 'Multiple groups'
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,6 @@ services:
       context: .
       dockerfile: Dockerfile-app
     restart: always
-    image: "iit-backend/app"
     ports:
       - "5000:5000"
     env_file:


docker-compose.yml
```diff
@@ -6,7 +6,6 @@ services:
       context: .
       dockerfile: Dockerfile-app
     restart: always
-    image: "iit-backend/app"
```
atatmaja, line 9: we don't want to actually push this change so don't forget to revert this

app/utils/estimate_prioritization/prioritization_criteria.py
```diff
@@ -7,8 +7,8 @@
     ],
     'adjustment_testadj': [
         lambda estimate: (estimate['pop_adj'] is True) and (estimate['test_adj'] is True),
-        lambda estimate: (pd.isna(estimate['pop_adj'])) and (estimate['test_adj'] is True),
         lambda estimate: (estimate['pop_adj'] is True) and (pd.isna(estimate['test_adj'])),
+        lambda estimate: (pd.isna(estimate['pop_adj'])) and (estimate['test_adj'] is True),
```
annoyingpr, line 11: nit: try to avoid changes that just move stuff around but don't change the actual code

atatmaja: This change is actually functional @annoyingpr. The estimate prioritization logic goes thru the prio criteria defined in this array one by one and applies them on a DF of estimates. The order in which the prioritization criteria are defined is important - thus the change in ordering is actually a functional change.

annoyingpr: oh i see now! good to know, my bad @123chrisc 😅

rahularoradfs, line 11: Seems fair to me @123chrisc - could certainly make a cogent argument that now with higher seroprevalence and better assays, we should prioritize pop-adj test-unadj in these cases

app/utils/estimate_prioritization/prioritization_criteria.py
```diff
@@ -53,8 +52,11 @@
     ],
     'test_type': [
         lambda estimate: estimate['test_type'] == 'Neutralization',
+        lambda estimate: estimate['test_type'] == 'Multiple Types',
+        lambda estimate: estimate['test_type'] == 'ELISA',
         lambda estimate: estimate['test_type'] == 'CLIA',
-        lambda estimate: estimate['test_type'] == 'ELISA'
+        lambda estimate: estimate['test_type'] == 'LFIA',
+        lambda estimate: estimate['test_type'] == 'Other',
```
rahularoradfs, line 59: These 'catch-call' cases are already implied to be the last thing in the hierarchy, if they're not listed among the options  listed- but it's good to be explicit about them!

docker-compose.yml
rahularoradfs: @123chrisc

rahularoradfs (commented):
@123chrisc, congratulations on the pull request!! Overall looks good to me. I do have a few broader questions before we go ahead and approve this.

Things that should be done here:
- Did you check to ensure that all variable names are correct here?

Things that don't necessarily need to be part of this PR, but need to be done by the time we get the meta-analysis in? 
- Did you check to ensure that all variables here map to the correct variables in the Airtable? This is a check we should do in general to ensure that all Airtable changes have actually been propagated to the database. Here is the file you need to check to ensure that: https://github.com/serotracker/iit-backend/blob/master/app/utils/airtable_fields_config.py . In this file, focus on the `full_airtable_fields` mapping; things before colon are Airtable names from estimates field; after colon are names in the database
- Have you checked to ensure that every variable has a pooling function in https://github.com/serotracker/iit-backend/blob/master/app/utils/estimate_prioritization/pooling_functions.py ?

Would be great if you could share a plan for the last two :)

app/utils/estimate_prioritization/prioritization_criteria.py
123chrisc: Variables were all checked to match. Sorting pooling functions now!

docker-compose.yml
123chrisc: Sorry for late follow-up on this. Did not open GitHub for a hot minute. Is this change sorted or do I need to go back and reverse it?

app/utils/estimate_prioritization/prioritization_criteria.py
123chrisc: All variables do have a pooling function. They are also matched to the airtable_fields_config.py

rahularoradfs:
Can we get this merged @123chrisc @simonarocco?

simonarocco:
> Can we get this merged @123chrisc @simonarocco?

@123chrisc just an FYI, as the PR request owner, it is your responsibility to get the PR merged. this means getting the approvals that you need to merge the PR (right now you have zero approvals). once you get these approvals you can merge. i believe there are still some unaddressed comments. once you have addressed a reviewer's comment please "resolve" the comment

docker-compose.yml
simonarocco: @123chrisc you need to undo this change, i.e. put this line back please.

simonarocco (changes requested):
this looks good to me, but you need to revert the change to the docker-compose.yml file. also, please get rahul's approval on the changes to estimate prioritization before merging this

PR #270, opened 2021-10-01
Write data back to airtable from ETL

simonarocco:
- This PR includes a new function called batch_update_records which writes new field values for records back to Airtable
- This uses the pyairtable Python client
- We must make one request per record change as the Airtable API batch updating functionality is broken
- This function is generalizable such that we can later right other fields back to airtable too such as case counts and vaccines
- In testing this PR I wrote all the test adj data we currently have in our DB back to airtable


app/database_etl/airtable_records_handler/airtable_records_formatter.py
```diff
@@ -105,3 +120,21 @@ def apply_study_max_estimate_grade(df: pd.DataFrame) -> pd.DataFrame:
...
+
+
+def batch_update_airtable_records(records_to_update, field_names):
+    AIRTABLE_API_KEY = os.getenv('AIRTABLE_API_KEY')
+    AIRTABLE_BASE_ID = os.getenv('AIRTABLE_BASE_ID')
+
+    # Create table object
+    table = Table(AIRTABLE_API_KEY, AIRTABLE_BASE_ID, 'Rapid Review: Estimates')
+
+    # Cycle through all the records to update
```
rahularoradfs, line 132: `iterrows` broadly recommended against for perf reasons. What I might suggest is turning the inside of the loop into a function, and then using `df.apply`. See relevant answer here: https://stackoverflow.com/questions/24870953/does-pandas-iterrows-have-performance-issues

app/database_etl/airtable_records_handler/airtable_records_formatter.py
```diff
@@ -105,3 +120,21 @@ def apply_study_max_estimate_grade(df: pd.DataFrame) -> pd.DataFrame:
...
+    for i, row in records_to_update.iterrows():
+        # For each record, create a dict where the key is the field name and the value is the new field value
+        # If the value is NaN, convert to None because NaN throws error with Airtable API
+        fields = {x: row[airtable_fields_config[x]] if not pd.isna(row[airtable_fields_config[x]]) else None for x in field_names}
```
rahularoradfs, line 136: Might be helpful to comment what exactly is going on in this line. We're taking all the field names in Airtable and then mapping them to their values, but we need to get their values by indexing based on their database names... wonder if there's a way to rewrite in terms of database_names and an inverse mapping of airtable_fields_config that is more intuitive, or if a comment would help? had to stare at this for a couple mins to understand

rahularoradfs (commented):
this just adds the functinality to update records, right - and doesn't automatically do that for test adjustment yet? if so, this looks solid to me, conditional on this being the approach we want to take.

The alternative approach here is the AirTable automations approach, which is more consistent with ultimately simplifying the ETL. This would involve setting up automations for eg cases per 100 such that when a country name and a sampling set of dates are filled in, the AT automation triggers a microservice/lambda that provides the required data. I'd lean a bit more in this direction long term, tbh. Also recognize the long term and short term tradeoffs

Approval should come from a more experienced dev!

app/database_etl/airtable_records_handler/airtable_records_formatter.py
simonarocco: I believe df.apply works if we are applying a function to a df and we want the output to be within the existing df. That's why the function returns the df with the function applied. I think that's what vectorization typically allows you to do - apply a transformation to a matrix as an entire operation instead of going row by row. In this case, we don't care about generating an output and storing it in the original records_to_update df. We want to cycle through the df and create a dictionary from it. Not sure that df.apply makes sense in this case for that reason. I do know there are performance issues, which is why vectorization is preferred when you are applying functions to the df across all rows and cols.

app/database_etl/airtable_records_handler/airtable_records_formatter.py
simonarocco: Ya sure can comment. I can't really make the code more readable by extracting out variables unless I turn it into a second for loop without the dict comprehension.

simonarocco:
> this just adds the functinality to update records, right - and doesn't automatically do that for test adjustment yet? if so, this looks solid to me, conditional on this being the approach we want to take.
> 
> The alternative approach here is the AirTable automations approach, which is more consistent with ultimately simplifying the ETL. This would involve setting up automations for eg cases per 100 such that when a country name and a sampling set of dates are filled in, the AT automation triggers a microservice/lambda that provides the required data. I'd lean a bit more in this direction long term, tbh. Also recognize the long term and short term tradeoffs
> 
> Approval should come from a more experienced dev!

1. No, I'm changing this PR to actually use the function i've created to write test adjustment data back to airtable. I don't see why we wouldn't want to use it now (the point of me doing the PR was to get the test adj data back to airtable asap)

2. We've flip flopped back and forth between doing it this way and using the automation. I've spent many hours working with the automation and while it might be a good long term solution, I just don't really have the time to move everything over to making tiny microservices and endpoints and threading them into the automations. I also still am almost 100% sure the test adj endpoint still takes over 30 seconds so it would time out with the airtable automation.

app/database_etl/airtable_records_handler/airtable_records_formatter.py
rahularoradfs: `df.apply` returns a `pd.series` object where the values of the series are the outputs of the function specified, applied to each row of the input df. It doesn't necessarily need to involve storing the result in the original df. Instead of writing `records_to_update['newcol'] = records_to_update.apply(column_creation_function, axis = 1)`, you can very much write `records_to_update.apply(function_with_side_effects, axis = 1)` as a line on its own that triggers the desired function with the desired side effects for each row in the df.

rahularoradfs (commented):
replied to comments

rahularoradfs:
> 
> 
> > this just adds the functinality to update records, right - and doesn't automatically do that for test adjustment yet? if so, this looks solid to me, conditional on this being the approach we want to take.
> > The alternative approach here is the AirTable automations approach, which is more consistent with ultimately simplifying the ETL. This would involve setting up automations for eg cases per 100 such that when a country name and a sampling set of dates are filled in, the AT automation triggers a microservice/lambda that provides the required data. I'd lean a bit more in this direction long term, tbh. Also recognize the long term and short term tradeoffs
> > Approval should come from a more experienced dev!
> 
>     1. No, I'm changing this PR to actually use the function i've created to write test adjustment data back to airtable. I don't see why we wouldn't want to use it now (the point of me doing the PR was to get the test adj data back to airtable asap)
> 
>     2. We've flip flopped back and forth between doing it this way and using the automation. I've spent many hours working with the automation and while it might be a good long term solution, I just don't really have the time to move everything over to making tiny microservices and endpoints and threading them into the automations. I also still am almost 100% sure the test adj endpoint still takes over 30 seconds so it would time out with the airtable automation.

Got it re 1. Just didn't see it in the version I reviewed and wanted to make sure I wasn't missing anything.

Re 2: ultimately at your discretion. Test adj running quit fast locally for me with the proper parameters, making sure we are passing in reasonable input data, etc - definitely ~1s per record or less to run the model on my local machine. But remains up to you.

app/database_etl/airtable_records_handler/airtable_records_formatter.py
atatmaja: Agree with rahul here. 

Take this with a grain of salt but I found an article comparing a bunch of ways to apply a function over a DF (https://towardsdatascience.com/apply-function-to-pandas-dataframe-rows-76df74165ee4), apparently itertuples and list comp perform better than apply??

app/config.py
```diff
@@ -58,6 +59,7 @@ class ApiProductionConfig(ApiConfig):
     DATABASE_PASSWORD = os.getenv('DATABASE_PASSWORD')
     DATABASE_HOST_ADDRESS = os.getenv('DATABASE_HOST_ADDRESS')
     DATABASE_NAME = 'whiteclaw'
+    WRITE_TO_AIRTABLE = True
```
atatmaja, line 62: I think it would be better to encode this as an environment variable in addition to tying it to the prod config. Reason being that it might be easy for devs to make the mistake of setting FLASK_ENV=prod.

```suggestion
    WRITE_TO_AIRTABLE = os.getenv('WRITE_TO_AIRTABLE', False)
```

requirements.txt
```diff
@@ -126,7 +127,6 @@ requests-oauthlib==1.3.0
 requests-toolbelt==0.9.1
 rsa==4.7.2
 s3transfer==0.3.4
-scipy==1.5.0
```
atatmaja, line 129: Do we need scipy as a dependency for something?

atatmaja:
I love the automations solution in theory, but on second thought it's a bit too brittle for me to be comfortable with in a production pipeline. Specifically I'm concerned about:
- The automation run limit
- The lack of visibility into when our automations aren't running properly
- The lack of built in failure recovery (i.e. ability to automatically retry testadj computation/writeback upon failure)

For now i think what simona has is a good solution. I would personally like to see more monitoring before we can run this in production though (i.e. airtable IDs of the records we were supposed to write to, which ones actually got written to and which ones didn't - basically stuff that will help us to diagnose issues and be able to manually resolve them if absolutely necessary)

app/database_etl/airtable_records_handler/airtable_records_formatter.py
rahularoradfs: @atatmaja seems to vary by use case / function called and optimizability. See https://towardsdatascience.com/how-to-make-your-pandas-loop-71-803-times-faster-805030df4f06

rahularoradfs:
Good points against automation, both.

requirements.txt
simonarocco: idk why this got deleted will readd

atatmaja (approved):
LGTM

PR #285, opened 2021-11-07
Adding in None as option for multi select tables

simonarocco:
- PR adds in functionality to preserve "None" as an option in multi select tables (city, state, test manufacturer, antibody target_
- Also changing filter options to only return Spike and Nucleocapsid for antibody target as request by @rahularoradfs

### DIFF
--- a/app/database_etl/airtable_records_handler/airtable_records_formatter.py
+++ b/app/database_etl/airtable_records_handler/airtable_records_formatter.py
@@ -64,6 +64,11 @@ def get_most_recent_publication_info(row: Dict) -> Dict:
     return row
 
 
+def replace_null_fields(row_val):
+    null_cols = ['nr', 'NR', 'Not Reported', 'Not reported', 'Not available', 'NA', 'N/A']
+    return [i if i not in null_cols else None for i in row_val]
+
+
 def standardize_airtable_data(df: pd.DataFrame) -> pd.DataFrame:
     # List of columns that are lookup fields and therefore only have one element in the list
     single_element_list_cols = ['included', 'source_name', 'url', 'source_publisher', 'summary',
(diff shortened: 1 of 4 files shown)


simonarocco:
@atatmaja @annoyingpr how will this PR affect the frontend? will it? basically, now, if the antibody target or test manufacturer is not reported, it will be "null" (i.e. an empty []) in the request body. currently we show "not reported" but this is not consistent with the rest of the fields for which we show "null" if there are no values. so, will returning [] for antibody target now break anything on frontend?

app/namespaces/data_provider/data_provider_service.py
```diff
@@ -201,7 +201,7 @@ def get_all_filter_options() -> Dict[str, Any]:
 
         # Get antibody target
         query = session.query(distinct(AntibodyTarget.antibody_target_name))
-        results = [q[0] for q in query if q[0] is not None]
+        results = [q[0] for q in query if q[0] is not None and q[0] in ['Spike', 'Nucleocapsid (N-protein)']]
```
atatmaja, line 204: nit: add comment with reasoning for why we don't want to surface these options

app/database_etl/airtable_records_handler/airtable_records_formatter.py
```diff
@@ -79,8 +84,12 @@ def standardize_airtable_data(df: pd.DataFrame) -> pd.DataFrame:
...
 
     # Convert elements that are "Not reported" or "Not Reported" or "NR" to None
-    df.replace({'nr': None, 'NR': None, 'Not Reported': None, 'Not reported': None,
-                'Not available': None, 'NA': None}, inplace=True)
+    df.replace({'nr': None, 'NR': None, 'Not Reported': None, 'Not reported': None, 'Not available': None,
+                'NA': None, 'N/A': None}, inplace=True)
+
+    # Antibody target reported are multi select and will have ['Not reported'], ['NR'] - convert to None
+    # Can also have a situation with ["Spike", "Not reported"] so need to check within list
+    df['antibody_target'] = df['antibody_target'].apply(lambda x: replace_null_fields(x))
```
atatmaja, line 92: shouldn't line 87 do what line 92 does? I'm a bit confused

atatmaja (approved):
Looks fine but a few clarifying questions

app/database_etl/airtable_records_handler/airtable_records_formatter.py
simonarocco: no it doesn't because the antibody target column is like: ["Not reported", "Spike], ["Nucleocapsid], ["Spike, Nucleocapsid"] so we need to look inside the lists within the rows to replace the Not reported

PR #286, opened 2021-11-11
Setting estimate prio bool to false

simonarocco:
1. This PR removes the /dashboard_records endpoint. The only endpoint available now is the /records endpoint
2. This PR removes the 'prioritize_estimates' bool param from the endpoint
3. This PR adds a parameter called estimates_subgroup that accepts three values: 'all'_estimates, 'prioritized_estimates', and 'primary_estimates'. The parameter is 'all' by default.

--> If 'all' is specified (or the param is left blank), ALL estimates will be returned (~13K estimates, multiple estimates per study)
--> If 'estimate_prioritization' is specified, only estimates output from estimate_prioritization will be returned
--> If 'primary_estimates', only the primary estimates of each study will be returned

- Will need coordinate deployment with frontend with @atatmaja and will need to coordinate updating the R package with @harrietware

### DIFF
--- a/app/namespaces/data_provider/data_provider_controller.py
+++ b/app/namespaces/data_provider/data_provider_controller.py
@@ -40,7 +40,7 @@ class Records(Resource):
 
         columns_requested = data.get('columns')
         research_fields = data.get('research_fields')
-        estimates_subgroup = data.get('estimates_subgroup', 'all')
+        estimates_subgroup = data.get('estimates_subgroup', 'all_estimates')
         prioritize_estimates_mode = data.get('prioritize_estimates_mode', 'dashboard')
         include_disputed_regions = data.get('include_disputed_regions', False)
         include_subgeography_estimates = data.get('include_subgeography_estimates', False)
@@ -122,7 +122,7 @@ class PaginatedRecords(Resource):
         reverse = data.get('reverse', None)
         columns = data.get('columns')
         research_fields = data.get('research_fields')
-        estimates_subgroup = data.get('estimates_subgroup', 'all')
+        estimates_subgroup = data.get('estimates_subgroup', 'all_estimates')
         prioritize_estimates_mode = data.get('prioritize_estimates_mode', 'dashboard')
         sampling_start_date, sampling_end_date = convert_start_end_dates(data, use_sampling_date=True)
         include_in_srma = data.get('include_in_srma', False)
--- a/app/namespaces/data_provider/data_provider_controller.py
+++ b/app/namespaces/data_provider/data_provider_controller.py
@@ -122,14 +122,15 @@ class PaginatedRecords(Resource):
         reverse = data.get('reverse', None)
         columns = data.get('columns')
         research_fields = data.get('research_fields')
-        prioritize_estimates = data.get('prioritize_estimates', True)
+        estimates_subgroup = data.get('estimates_subgroup', 'all')
         prioritize_estimates_mode = data.get('prioritize_estimates_mode', 'dashboard')
         sampling_start_date, sampling_end_date = convert_start_end_dates(data, use_sampling_date=True)
         include_in_srma = data.get('include_in_srma', False)
 
-        result = get_filtered_records(research_fields, filters, columns, sampling_start_date=sampling_start_date,
+        result = get_filtered_records(research_fields, filters, columns,
+                                      sampling_start_date=sampling_start_date,
                                       sampling_end_date=sampling_end_date,
-                                      prioritize_estimates=prioritize_estimates,
+                                      estimates_subgroup=estimates_subgroup,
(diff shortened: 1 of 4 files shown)


rahularoradfs:
@simonarocco This means that when you call the endpoint, you'll get several thousand data points back - more than one for each study. Are you sure this is what you want? How will you make sure to only show one for each study on the dashboard - will you be filtering by the `most_adjusted_primary_estimate` variable? Has the requisite front-end work been done for that?

atatmaja (changes requested):
As rahul commented, this will return all estimates back. For now we will want to return all primary estimates that match filters right?

rahularoradfs:
"just return primary estimates if estimate prioritization is false" - this won't work, as we have research use cases to want all estimates if estimate prio is false. Will need a separate primary estimates flag or filter

Separately, the 'dashboard_primary_estimate' variable - does this correspond to most adjusted primary estimate in AriTable?

app/namespaces/data_provider/data_provider_schema.py
```diff
@@ -10,7 +10,9 @@ class RecordsSchema(Schema):
     per_page = fields.Integer(allow_none=True)
     reverse = fields.Boolean(allow_none=True)
     research_fields = fields.Boolean(allow_none=True)
-    prioritize_estimates = fields.Boolean(allow_none=True)
+    estimates_subgroup = fields.String(validate=validate.OneOf(['all',
```
atatmaja, line 13: v nit but would prefer the following naming, either:
['all', 'prioritized', 'primary'] or ['all_estimates', 'prioritized'_estimates, 'primary_estimates']

simonarocco: sure

app/utils/get_filtered_records.py
```diff
@@ -221,10 +219,16 @@ def date_filter(record, start_date=None, end_date=None, use_sampling_date=True):
                 # True/False have dtype="object" instead of "bool"
                 if True in prioritized_records[col].values:
                     prioritized_records[col] = prioritized_records[col].fillna(False)
-            # Filling all NaN values with None: https://stackoverflow.com/questions/46283312/how-to-proceed-with-none-value-in-pandas-fillna
             prioritized_records = prioritized_records.fillna(np.nan).replace({np.nan: None})
         result = prioritized_records.to_dict('records')
 
+    # If estimates_subgroup is 'primary_estimates', just return the primary estimate for each study
```
atatmaja, line 225: I think this would be more performant to do this filtering step right after `get_all_records` is called (and before all other filtering happens)

atatmaja (commented):
Couple comments

app/utils/get_filtered_records.py
```diff
@@ -116,39 +116,59 @@ def get_all_records(research_fields=False, include_disputed_regions=False,
...
+                # Can't simply check the dtype because cols with True/None instead of
+                # True/False have dtype="object" instead of "bool"
+                if True in prioritized_records[col].values:
+                    prioritized_records[col] = prioritized_records[col].fillna(False)
+            prioritized_records = prioritized_records.fillna(np.nan).replace({np.nan: None})
+        query_dicts = prioritized_records.to_dict('records')
+
+    # If estimates_subgroup is 'primary_estimates', just return the primary estimate for each study
+    # Otherwise, estimates_subgroup = 'all' so just return all estimates
+    elif estimates_subgroup == 'primary_estimates':
```
atatmaja, line 167: Hmm I was only certain that this block would work well if moved before the filtering step and not estimate prioritization. Can you double check the results of this change for me, thanks.

simonarocco: well i didn't test the timing to see if it's more efficient to put here but in terms of results it still gives the accurate results when you move this block up

simonarocco: i.e. when "all_estimates" is selected it returns ~13k records, otherwise for the other 2 options it's about 3k records

simonarocco: all you are doing is doing all this logic right after get all records, converting query dicts to df, running this logic, then converting df back to query dicts and carrying on with a subset of the results

app/namespaces/data_provider/data_provider_controller.py
```diff
@@ -179,14 +122,15 @@ def post(self):
         reverse = data.get('reverse', None)
         columns = data.get('columns')
         research_fields = data.get('research_fields')
-        prioritize_estimates = data.get('prioritize_estimates', True)
+        estimates_subgroup = data.get('estimates_subgroup', 'all_estimates')
```
atatmaja, line 125: Should the default be "all_estimates" or "primary_estimates"? Given that the dashboard is our main client, I think it should be "primary_estimates"

atatmaja (approved):

PR #297, opened 2022-03-07
Adding in 2 new fields and modifying estimate_creator_csv

simonarocco:
- Adding in 'alpha_3_code' to the country table
- Adding in 'superseding_study' to the dashboard source table
- Modifying the etl_csv_creator script so it can be used to generate the Github csv and the biblio csv depending on the env var "CSV_TYPE"


atatmaja:
LGTM once you remove print statements. 

Also tests are failing.... probably just because they expect a certain response schema that we've since intentionally changed. We should either update our tests to reflect new schemas, refactor tests so that they're less stringent (i.e. just make sure that we get the right response code or something), or unlink our tests from CI entirely, cause right now we're just getting false signal about the validity of our code lol.

PR #309, opened 2023-01-24
Made it so the request is hit multiple times before failing to avoid …

HimanshuRanka:
…one off gateway timeouts

### DIFF
--- a/app/database_etl/airtable_records_handler/airtable_records_getter.py
+++ b/app/database_etl/airtable_records_handler/airtable_records_getter.py
@@ -29,10 +29,13 @@ def add_fields_to_url(url, fields):
 def airtable_get_request(url: String, headers):
     # Make request and retrieve records in json format
     r = requests.get(url, headers=headers)
-    if r.status_code != 200:
-        print(r.text)
+    i = 0
+    while r.status_code != 200 and i < 3:
         print(r.status_code)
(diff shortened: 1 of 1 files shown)


app/database_etl/airtable_records_handler/airtable_records_getter.py
```diff
@@ -29,10 +29,13 @@ def add_fields_to_url(url, fields):
...
     # Make request and retrieve records in json format
     r = requests.get(url, headers=headers)
-    if r.status_code != 200:
-        print(r.text)
+    i = 0
+    while r.status_code != 200 and i < 3:
         print(r.status_code)
-        print(r.json())
+        print(r.text)
+        r = requests.get(url, headers=headers)
```
Gurmie12, line 36: Can't we just pass a timeout into this? So that the request waits longer for a response

HimanshuRanka: The airtable endpoint is timing out. But what do u mean pass a timeout into this (as in into the requests.get?

Gurmie12 (approved):
LGTM! Should fix the issue but at some point when there's more bandwidth we should dig deeper into this

HimanshuRanka:
Yeah that makes sense. I will whip up a ticket to take a look

PR #311, opened 2023-02-03
Fix estimates csv generator script

Gurmie12:
estimates csv generator script was not outputing the `antibody_target` and `isotype` columns as expected (refer to ticket).
https://airtable.com/appT8tnLbGJV6v81V/tbli2lWQHAqBa6ZcI/viwFS5nE6gepmRwVB/rechvFUuXWyx4oaGA?blocks=hide
Ran the script again and the output was now showing as expected. Updated the public repo with the new CSV manually for the time-being

### DIFF
--- a/app/database_etl/airtable_records_handler/airtable_records_formatter.py
+++ b/app/database_etl/airtable_records_handler/airtable_records_formatter.py
@@ -71,6 +71,10 @@ def replace_null_fields(row_val):
     if row_val == None:
         return [None]
     null_cols = ['nr', 'NR', 'Not Reported', 'Not reported', 'Not available', 'NA', 'N/A']
+
+    if type(row_val) is str:
+        row_val = row_val.split(", ")
+
     return [i if i not in null_cols else None for i in row_val]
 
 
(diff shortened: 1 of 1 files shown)


rezahosseinimed:
Thanks, Gurman. I checked last week and the variables were fine then. I just checked again and "isotypes" and "antibody_target" have the same problem again.

Another issue with the "isotypes" and "antibody_target" variables is they have repeated groups. E.g. in "isotypes", we have "IgG, IgM" and "IgM, IgG" as two separate groups while they should be one. The problem originally comes from Airtable multiple-select fields where users can choose different options in different orders. There is a hacky way to sort the selected options of these two variables in Airtable but not automatic and optimal. Do you think we can sort the contents of these two variables in the backend instead?

HimanshuRanka (dismissed):

Gurmie12:
> Thanks, Gurman. I checked last week and the variables were fine then. I just checked again and "isotypes" and "antibody_target" have the same problem again.
> 
> Another issue with the "isotypes" and "antibody_target" variables is they have repeated groups. E.g. in "isotypes", we have "IgG, IgM" and "IgM, IgG" as two separate groups while they should be one. The problem originally comes from Airtable multiple-select fields where users can choose different options in different orders. There is a hacky way to sort the selected options of these two variables in Airtable but not automatic and optimal. Do you think we can sort the contents of these two variables in the backend instead?

The issue is persisting since this PR hasn't been merged into the master branch yet so the daily scheduled action is running on the old code, which should be resolved once this is merged. As for the sorting, I have pinged you on basecamp

HimanshuRanka (approved):

PR #325, opened 2023-06-14
updated endpoint

HimanshuRanka:
Updated the endpoints to return the most recent pub date as well as the last updated date


Gurmie12 (approved):

app/namespaces/data_provider/data_provider_controller.py
```diff
@@ -243,4 +243,5 @@ def get(self):
             args=dict(request.args)))
 
         result = get_all_filter_options()
+        print(result)
```
Gurmie12, line 246: Not sure if you want this but if it was just for testing maybe remove?

PR #327, opened 2023-07-09
Gurman/remove ind eval tab column

Gurmie12:
This PR addresses our ETL currently not running as a result of some columns being removed from Airtable. This PR removes the column from the ETL and also updates our database schema.

N/A

Ran the ETL locally and verified that the records were pushed successfully to my local database.

No, although, we will have to go in and manually apply the new database migrations.


RezaHss:
Hi ﻿Gurman﻿, thanks for the update!

I have encountered a new problem since yesterday when you pushed some changes. The problem is:

The [tableau\_analyze\_records.xlsx](https://drive.google.com/drive/folders/18DkAftv9jPOfShGW3OCo_SVm_KHyFg9L)  got updated but its first tab has 11 records compared to 23962 in the previous version. The 3rd tab has some changes too. This prevents us from updating the Tableau graphs.

app/namespaces/test_adjustment/test_adjustment_service.py
```diff
@@ -178,12 +178,12 @@ def get_adjusted_estimate(self, test_adj, ind_se, ind_sp, ind_se_n, ind_sp_n,
         if pd.isna(test_adj):
 
             # Independent evaluation is available
```
harrietware, line 180: Change the prioritization order so Author evaluation is first

harrietware (approved):
Looks good to me, note one comment to change the prioritization order!

app/namespaces/test_adjustment/test_adjustment_service.py
Gurmie12: [e424cf4](https://github.com/serotracker/iit-backend/pull/327/commits/e424cf4385419691d1e89338aa43680dffc0d1b2)

harrietware (approved):
lgtm!

Gurmie12:
> Hi ﻿Gurman﻿, thanks for the update!
> 
> I have encountered a new problem since yesterday when you pushed some changes. The problem is:
> 
> The [tableau_analyze_records.xlsx](https://drive.google.com/drive/folders/18DkAftv9jPOfShGW3OCo_SVm_KHyFg9L) got updated but its first tab has 11 records compared to 23962 in the previous version. The 3rd tab has some changes too. This prevents us from updating the Tableau graphs.

Already answered in DM but will answer here too just for visibility to others. There were only 11 records as it was run against a small subset of the data for testing but will go back to the expected number of records when we merge into master and run against production!

app/database_etl/tableau_data_connector/google_services_manager.py
```diff
@@ -19,8 +19,8 @@ def __init__(self, sheet_name):
 
     def create_service(self):
         creds = None
-        if os.path.exists('tableau_data_connector/token.pickle'):
-            with open('tableau_data_connector/token.pickle', 'rb') as token:
+        if os.path.exists('app/database_etl/tableau_data_connector/credentials.json'):
```
HimanshuRanka, line 22: this checks if credentials.json exists and then uses token.pickle below is that expected?

HimanshuRanka (approved):

app/database_etl/tableau_data_connector/google_services_manager.py
Gurmie12: Nice Catch!

Gurmie12: [80ad6f4](https://github.com/serotracker/iit-backend/pull/327/commits/80ad6f490dfd5245982593a4578ca9273106bcdc)

PR #333, opened 2023-11-26
Improved quality of latitude and longitude data used for arbotracker estimate data.

SeanKennyNF:

Addresses https://github.com/PathoTracker/Pathotracker/issues/28

N/A

I've run a similar version of this script locally and couldn't find pins in the incorrect countries anymore. I looked at lots of the edge cases (Ex. Tyrol in Austria and the Andoman and Nicobar islands in Austria where the state name is given as a city name and requires a city lookup and a fallback to state) as well as . I've verified that no null latitudes and longitudes are present after my change.

No


Pathogens/Utility/location_utils/country_iso2.json
```diff
@@ -198,6 +198,7 @@
   "Slovakia": "SK",
   "Slovenia": "SI",
   "Solomon Islands": "SB",
+  "Solomon islands": "SB",
```
SeanKennyNF, line 201: Airtable data has inconsistent capitalization for this country.

Pathogens/Utility/location_utils/country_iso2.json
```diff
@@ -210,6 +211,7 @@
   "Switzerland": "CH",
   "Syrian Arab Republic": "SY",
   "Tajikistan": "TJ",
+  "Taiwan": "TW",
```
SeanKennyNF, line 214: Contained in airtable but not in country code files, you can check Wikipedia if you want to make sure I'm not making these up lol

Pathogens/Utility/location_utils/location_functions.py
```diff
@@ -178,21 +178,51 @@ def compute_pin_info(df: pd.DataFrame, geo_dfs: dict) -> pd.DataFrame:
...
+def parse_mapbox_response(response):
+    data = response.json()
+    if data and "features" in data and len(data['features']) > 0:
+        return data['features'][0]['center']
+    else:
         return None
 
-    url = f"https://api.mapbox.com/geocoding/v5/mapbox.places/{geocoder_search_text}.json?" \
-          f"access_token={os.getenv('MAPBOX_API_KEY')}&types={geocoder_data_type}"
+def get_city_lat_lng(city_name, state_name, country_name):
```
SeanKennyNF, line 192: Kind of confusing but the story of this file is basically that I set up like a whole bunch of fallbacks because I kept finding these edge cases in our data where the city name would be the name of the state or the city name would be there and the state name wouldn't be there, etc.

With all of these fallbacks literally every pin has a lat and lng and I did a really good search through the pins to see if I could find any in the wrong country and I only found one which was on the wrong island in Indonesia but that's literally it so I'm pretty confident this is good. My only thought for a future enhancement here would maybe be an override json object where we can put specific locations where we know mapbox has bad data for (so far just literally this one spot in Indonesia) and then correct the latlngs ourselves but I can take that on later, let's just fix 99.9% of these first :)

HimanshuRanka: Ok sounds good! And also just a note on being in the wrong island, that is fine we only really care about country level correctness AFAIK and will probably have to introduce some noise on purpose to ensure all pins are visible (multiple pins in the same city would overlap otherwise)

Pathogens/Utility/location_utils/location_functions.py
```diff
@@ -178,21 +178,51 @@ def compute_pin_info(df: pd.DataFrame, geo_dfs: dict) -> pd.DataFrame:
...
+    if(coords is None):
+        return get_country_lat_lng(country_name)
+    else:
+        return coords
+
+def get_country_lat_lng(country_name):
+    url = get_mapbox_api_query_url(country_name, 'country', country_code=get_country_code(country_name=country_name, iso3=False))
+
+    api_response = requests.get(url)
+    coords = parse_mapbox_response(api_response)
```
HimanshuRanka, line 227: I think one of the things we probably want to consider here is if mapbox allows batch requests and if we can do that so instead of having 1102 requests we have at max 3.

Pathogens/Arbo/ETL/arbo_etl_main.py
```diff
@@ -133,20 +133,14 @@ def main():
...
-                                                                 'place'), dtype='object')
-                                                             if pd.notnull(row['city']) and pd.notnull(row['state'])
-                                                             else (pd.Series(
-                                                                 get_lng_lat(row['state'], 'region'),
-                                                                 dtype='object') if pd.notnull(
-                                                                 row['state'])
-                                                                   else pd.Series(
-                                                                 get_lng_lat(row['country'], 'country'),
-                                                                 dtype='object')), axis=1)
+        pd.Series(get_city_lat_lng(
```
HimanshuRanka, line 138: I think it would be valuable to introduce a jitter pins functionality here this time around itself instead of waiting for the future. 

Would you be able to do that? essentially add or subtract up to 5 degrees from the Latitude and longitude of each estimate so they are all a little more scattered. 

Can reference the jitter_pins function if you need it!

HimanshuRanka (changes requested):
LGTM overall except the small jitter pins addition (aware it was not in the og ticket but now that I think of it, there is value in just getting it out of the way this time around)

Pathogens/Arbo/ETL/arbo_etl_main.py
SeanKennyNF: Yeah this is a sick idea. Definitely necessary, I didn't realize how many cities had a whole bunch of studies :+1:

I tried some values for what the maximum jitter should be and I really really liked the way the map looked on 0.1. When I turned the jitter up too high I didn't like what it did to the map but I definitely get that these things are subjective so if you want more jitter let's turn it up :chart_with_upwards_trend: but if you also like 0.1 let's send it :mailbox:

HimanshuRanka (approved):

PR #334, opened 2023-11-26
Estimates with sample sizes of under 5 are no longer saved in the database

SeanKennyNF:

Addresses https://github.com/PathoTracker/Pathotracker/issues/27

N/A

I've run a similar version of this script locally and couldn't find any estimates with a sample size of under 5 after my changes. The front end looks fine after my changes as well.

No


Pathogens/Arbo/ETL/arbo_etl_main.py
```diff
@@ -132,6 +133,9 @@ def main():
 
     # print(records_df.dtypes)
 
+    print("[STEP] filtering out studies that don't meet the minimum sample size requirement (sample sizes must be >={})".format(MINIMUM_SAMPLE_SIZE))
+    records_df.drop(records_df[records_df.sample_size < MINIMUM_SAMPLE_SIZE].index, inplace=True)
```
SeanKennyNF, line 137: Same deal as the location one. If you're not convinced this works let me know and I'll send the API data that you can plug into your front end to double check if you're interested.

PR #336, opened 2023-11-30
Added `estimate_id`, `who_region`, and `source_sheet_name` to the Arbo database schema and ETL.

SeanKennyNF:
## Briefly describe the feature or bug that this PR addresses.

Added `estimate_id`, `who_region`, and `source_sheet_name` to the Arbo database and ETL.

## Please link the Airtable ticket associated with this PR.

https://github.com/PathoTracker/Pathotracker/issues/38

## Describe the steps you took to test the feature/bugfix introduced by this PR.

TODO

## Does any infrastructure work need to be done before this PR can be pushed to production?

No

PR #338, opened 2023-11-30
Fix Arbo ETL

SeanKennyNF:

When we merged https://github.com/serotracker/iit-backend/pull/336, I accidentally forgot to commit this line so the Arbo ETL was not working. This PR is the fix we made in prod to make things work again

https://github.com/PathoTracker/Pathotracker/issues/38

My testing for my previous PR included this change so everything carries over. We also ran the ETL with this change in production, I'm just merging it so that the next person doesn't have to deal with the headache.

No

### DIFF
--- a/Pathogens/Arbo/ETL/constants.py
+++ b/Pathogens/Arbo/ETL/constants.py
@@ -50,4 +50,4 @@ required_fields = [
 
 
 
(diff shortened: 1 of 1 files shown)


HimanshuRanka (approved):
whoops missed this

PR #341, opened 2023-12-11
Allow correction of locations whose format doesn't match what Mapbox expects (Fixes West Bengal, Bali, and a few others)

SeanKennyNF:

There's a summary comment but basically I'm adding a json file with overrides for specific locations that mapbox struggles with to fix a dozen or so locations on our map.

https://github.com/PathoTracker/Pathotracker/issues/84

I've checked every pin whose location I've adjusted and they are in a more correct location

Ex.
image
image
image

No


Pathogens/Utility/location_utils/city_state_country_mapbox_response_data_cache.py
```diff
@@ -0,0 +1,15 @@
+city_state_country_mapbox_response_data_cache = {}
```
SeanKennyNF, line 1: Same file as `city_state_country_lat_lng_data_cache.py`, I just renamed files because I'm storing more than just latlng values now.

Pathogens/Utility/location_utils/country_codes.py
```diff
@@ -0,0 +1,19 @@
+import os
```
SeanKennyNF, line 1: Sorry this is not new, moved from `location_functions.py` because that file was getting hard to navigate (at least for me)

Pathogens/Utility/location_utils/location_functions.py
```diff
@@ -178,70 +159,31 @@ def compute_pin_info(df: pd.DataFrame, geo_dfs: dict) -> pd.DataFrame:
 
     return df
 
-def get_mapbox_api_query_url(geocoder_search_text, geocoder_data_type, country_code):
```
SeanKennyNF, line 181: I didn't actually change any logic in this file, I just cleaned some of the spaghetti I wrote to get this going and moved it to `mapbox_api_client.py` in a less messy way.

Pathogens/Utility/location_utils/mapbox_api_client.py
```diff
@@ -0,0 +1,89 @@
...
+from Pathogens.Utility.location_utils.city_state_country_mapbox_response_data_cache import update_mapbox_response_cache, attempt_to_fetch_mapbox_response_data_from_cache
+from Pathogens.Utility.location_utils.country_codes import get_country_code, read_from_json
+
+MAPBOX_REQUEST_PARAM_OVERRIDES = read_from_json('mapbox_request_param_overrides.json')
+
+@dataclass
+class MapboxResponse:
+    center_coordinates: [float, float]
+    bounding_box: [float, float, float, float]
+    text: str
```
SeanKennyNF, line 14: Okay I'm not technically using `bounding_box` and `text` but I swear they are so helpful for debugging these kinds of problems and these fields are the only reason that I was able to find as many mistakes in out latlng stuff as I did so I'd really like to keep these around for the next time I need to debug this kind of problem please :sob:

Pathogens/Utility/location_utils/mapbox_api_client.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+
+    return MapboxRequestParams(
+        mapbox_search_text = city_name + "," + state_name,
+            country_code = country_code,
+        geocoder_data_type = 'place'
+    )
+
+def make_mapbox_request(city_name, state_name, country_name):
+    cached_query_value = attempt_to_fetch_mapbox_response_data_from_cache(city_name, state_name, country_name)
```
SeanKennyNF, line 79: If you're thinking that the cache is efficient when the API returns a response but is actually inefficient when the API returns `None`, you're correct and saw what I didn't see when I was writing this first.

I'll come back and make the cache more efficient in these cases in a separate PR :)

Pathogens/Utility/location_utils/mapbox_request_param_overrides.json
```diff
@@ -0,0 +1,154 @@
+{
```
SeanKennyNF, line 1: Q: Sean where on Earth did you get these values from?

A: I wrote some code to search for mapbox cities that aren't a part of the bounding box of their state. Basically if you were looking for "Calgary, Alberta", I would do the query for "Calgary, Alberta" and check whether the city returned actually belongs to the bounding box of "Alberta" or not. I got this list of 20 mapbox locations whose queries were returning something outside the correct state.
```
Akmola,Zhambyl Oblysy,Kazakhstan
Almaty,Qaraghandy Oblysy,Kazakhstan
Baganangone; Boguera; Gambo; Mabo,Lobaye,  Mbomou, Haut-Mbomou,Central African Republic
Bali,Bali,Indonesia
Bangkok,Krung Thep Maha Nakhon,Thailand
Birbhum,West Bengal,India
Costa Grande and Costa Chica,Guerrero,Mexico
Guerrero State,Guerrero,Mexico
Guerrero,Guerrero,Mexico
Gulu,Gulu,Uganda
Ham Hiep,Tỉnh Bình Thuận,Viet Nam
Ham Kiem, Tỉnh Bình Thuận,Viet Nam
Hooghly,West Bengal,India
Juruti,Pará,Brazil
Kafr Ayoub,Ash Sharqia Governorate,Egypt
Kampala,Kampala,Uganda
Karachi,Sindh, Punjab,Pakistan
Madhya Pradesh,Madhya Pradesh,India
Mbale,Mbale,Uganda
Mbarara,Mbarara,Uganda
Midnapur (West),West Bengal,India
North 24 Parganas,West Bengal,India
Okwoga,Benue,Nigeria
Osun (Coord.), Lagos,Osun,Nigeria
Osun,Osun,Nigeria
Port Moresby ,National Capital,Papua New Guinea
South 24 Parganas,West Bengal,India
São Francisco Valley,Bahia, Pernambuco,Brazil
West Bengal,West Bengal,India
```
The commonality here is that either
A. The city exists in a different state than the one specified (Ex. `Almaty,Qaraghandy Oblysy,Kazakhstan`)
B. The city is not a city but is a district within a state which is treated differently in mapbox (Ex. `North 24 Parganas,West Bengal,India`)
C. The city name is the name of a state and mapbox gets confused and returns some random city because it doesn't know what to do (Ex. `Madhya Pradesh,Madhya Pradesh,India`)
D. The city goes by a different name in mapbox (Ex. `Midnapur (West),West Bengal,India`)

I went through each of these and checked whether the location in prod matched what the correct location is and if it didn't, I made an entry in here with params that result in mapbox giving the correct location. If a city I listed doesn't appear in the list below, it's because I either messed up or I think the location in production is correct.

hmu if you have any questions :)

Pathogens/Utility/location_utils/mapbox_api_client.py
```diff
@@ -0,0 +1,93 @@
...
+    mapbox_request_param_override = MAPBOX_REQUEST_PARAM_OVERRIDES.get(country_code, {}) \
+        .get(state_name.strip() if state_name is not None else 'N/A', {}) \
+        .get(city_name.strip() if city_name is not None else 'N/A', None)
+
+    if(mapbox_request_param_override is not None):
+        return MapboxRequestParams(
+            mapbox_search_text = mapbox_request_param_override['mapbox_search_text'],
+            country_code = mapbox_request_param_override['country_code'],
+            geocoder_data_type = mapbox_request_param_override['geocoder_data_type']
+        )
```
SeanKennyNF, line 49: Meat and potatoes alert ⚠️

Basically takes the wheel away from my code for specific data points which I know are in a format that mapbox won't like.

Pathogens/Utility/location_utils/mapbox_api_client.py
HimanshuRanka: Just make a dev comment of this in the code so we do not accidentally remove it in another feature PR or code change

Pathogens/Utility/location_utils/mapbox_request_param_overrides.json
HimanshuRanka: Although I strongly appreciate the work it took to do this, I feel like this may be an issue that needs to be solved on a data level. I am OK to keep this in for now (if you agree) but would prefer to instead maybe update the ETL so that it generated a report that found faulty data (not sure if this was all done manually, and if so kudos to you! I think it would be an interesting problem to solve on checking if the city state and country values are valid or not. We can then pass this report onto the data team and have them make changes to the data as needed. Once the changes are made we can run the etl again. 

Let me know what your thoughts on this are.

In a situation where these happened for specific edge cases (and few) I would be OK hardcoding a solution but this is not maintainable in the long run and should not be on us as dev's to be fixing dirty data in a way that is extra complex. 

Ideally we never want to have an "overrides" file. I would not even be opposed to looking for an alternative API but since this works for 90% of our data and fails on some, it probably is an issue with the data and not the API.

HimanshuRanka (approved):
LGTM code wise. Left a couple comments about what I think we can do differently and a request to add in a comment for the unused fields you want to keep in.

Pathogens/Utility/location_utils/mapbox_request_param_overrides.json
SeanKennyNF: Everything you've written makes sense to me.

I gave some thought to the idea of a report that autogenerates and lists all the problematic data in the ETL and based on my understanding of the mapbox API that sounds viable to me so I'll give it a go and see where we get :)

SeanKennyNF: Alright my man, one report that generates possible faulty lat lng data for us to analyze and bring to the data team for them to change in Airtable coming right up :man_cook: https://github.com/serotracker/iit-backend/pull/342

I know that PR is a bit confusing, I tried to annotate it to the best of my ability but some of it kind of feels like rambles with not enough explanation which I'm sorry for but it's hard to summarize all of the warning signs that the data is bad, all of the different ways mapbox gives it's data, and how I believe the mapbox API is working. If you don't understand any part of that, let's hop on a call, I'm super down to talk about this there's just a lot going on and I need to know what part people are getting stuck on :)

SeanKennyNF: Great idea with the report btw :+1:

SeanKennyNF:
Closed in favour of https://github.com/serotracker/iit-backend/pull/342

PR #342, opened 2023-12-14
Added `lat_lng_generation_report.py` which records possible faults in our city and state data so that we can make changes to our city and state names in Airtable.

SeanKennyNF:

Please see the following thread: https://github.com/serotracker/iit-backend/pull/341#discussion_r1423118003
All this PR changes in terms of functionality is that we generate this report that has suggestions for a dev to look into and ask the data team to make in Airtable. The original PR has a lot more context that I won't include here because I don't want to repeat myself too much hahahahaha https://github.com/serotracker/iit-backend/pull/341

This PR might be a bit diffucult to understand so I've attached the report it generates here. It might help you understand what's going on or get into my head a bit better here.
[lat_lng_generation_report.txt](https://github.com/serotracker/iit-backend/files/13667407/lat_lng_generation_report.txt)

https://github.com/PathoTracker/Pathotracker/issues/84

Tested with front end pointed at the backend after I ran the ETL, nothing actually changed, the report was just generated which is the intended effect.

No


Pathogens/Utility/location_utils/mapbox_response_cache.py
```diff
@@ -0,0 +1,23 @@
+city_state_country_mapbox_response_cache = {}
```
SeanKennyNF, line 1: Couple changes to the cache

1. The cache works with the request params instead of `city_name`, `state_name`, and `country_name` which is important because I want to log the API requests for `district` and `place` seperately. It never mattered until now because we never made two seperate requests with those three parameters the same but it does matter now.
2. The cache saves "N/A" instead of None when the request is made but it returns no data. This was done because I want the cache to save failed requests as well as successful ones so that failed requests don't happen again and we just return a cached failure.
3. The cache saves the entire mapbox response instead of just the center coordinates.

Pathogens/Utility/location_utils/mapbox_api_client.py
```diff
@@ -0,0 +1,128 @@
...
+        if(mapbox_request_param_override is None) else mapbox_request_param_override
+
+    cached_query_value = attempt_to_fetch_mapbox_response_from_cache(mapbox_request_params)
+
+    # Important: The cache returns a value of "N/A" instead of None if the API request was attempted and it returned no
+    # data. What None means is that the response had never been cached in the first place. If you make all of these N/A's
+    # into None, if an API request returns no data, the entry in the cache won't be recognized as valid so we will make the
+    # request again even though we already know it won't return any data.
+    if(cached_query_value == "N/A"):
+        return None
```
SeanKennyNF, line 106: Code comment should be self explanatory but relates to a comment I left on the last file of this PR.

Pathogens/Utility/location_utils/mapbox_api_client.py
```diff
@@ -0,0 +1,128 @@
...
+            country_code = country_code,
+        geocoder_data_type = 'place'
+    )
+
+def make_mapbox_request(
+    city_name: str | None,\
+    state_name: str | None,\
+    country_name: str,\
+    log_file: TextIOWrapper | None,\
+    mapbox_request_param_override: MapboxRequestParams | None\
```
SeanKennyNF, line 94: I need this override because I want the report to be able to make requests with unusual parameters like giving a city name as a state name and giving a `geocoder_data_type` of `district` to see if that gives better results. This parameter isn't used otherwise.

Pathogens/Utility/location_utils/location_functions.py
```diff
@@ -1,27 +1,16 @@
 import os
+from io import TextIOWrapper
```
SeanKennyNF, line 2: I basically just did a cleanup of this file, I wouldn't worry too too much about the diff it's just moving stuff to different files really.

Pathogens/Utility/location_utils/lat_lng_generation_report.py
```diff
@@ -0,0 +1,314 @@
+from enum import Enum
```
SeanKennyNF, line 1: Alright this file is a bit hard to explain but I'll try. If you're confused at all it straight up might be better to talk in person so just send me a DM and we can hop on a call or something.

All of my `generate_line`-type functions with the exception of `generate_line_for_response_and_request` print out a line if they flag something that I've seen before that I know can be an issue for our data generation.

I know this function `WARN`s on a fair amount of false positives for bad data. My idea here is to autogenerate the report every so often and skim it for anything that seems serious and if I think it seems serious enough we go in and fix it. It's worth noting that almost every data point that actually causes a problem actually logs an `ERROR` and the `ERROR`s are almost all true positives for bad data.

I'm going to try to annotate this the best I can but sorry if it's still confusing despite that. Again, I'm super open to questions it's just a bit hard to explain after how far I've gone down the rabbit hole lol

I feel like reading one of the reports and seeing if you can follow along might help
[lat_lng_generation_report.txt](https://github.com/serotracker/iit-backend/files/13667486/lat_lng_generation_report.txt)

I think there's room to iterate on this but I'm pretty happy with how it turned out all things considered.

Pathogens/Utility/location_utils/lat_lng_generation_report.py
```diff
@@ -0,0 +1,314 @@
...
+        MapboxApiRequestLogLevel.WARN: "[WARN]",
+        MapboxApiRequestLogLevel.ERROR: "[ERROR]",
+    }
+    city_state_country_printout = "[" + (city_name if city_name is not None else 'N/A') + ", " + \
+        (state_name if state_name is not None else 'N/A') + ", " + country_name + "]"
+
+
+    return string_log_level_printout_map[logLevel] + " - " + city_state_country_printout
+
+def format_mapbox_text_and_matching_text_for_display(mapbox_response):
```
SeanKennyNF, line 40: Q: Sean what's all this `text` + `matching_text` garbage about?

A: Yeah so basically `text` can occasionally be in the native language of that country, the big one is that most of the ones in Egypt and Sudan seem to be in Arabic. What I found was that in situations like those `matching_text` is the English localized version but `matching_text` isn't always there so I'm just displaying both and the human that actually reads the report won't get stuck with just a name in Arabic or something and will always have as much information as possible to make their decision about whether or not this location is broken. I think this is okay, one of them was always localized to English from my experience.

There's a couple examples in that text file I linked.

Pathogens/Utility/location_utils/mapbox_api_client.py
```diff
@@ -0,0 +1,128 @@
...
+
+    if(region_data_in_context is None):
+        return None
+
+    return region_data_in_context.get('text', None)
+
+def parse_mapbox_response(response):
+    data = response.json()
+    if data and "features" in data and len(data['features']) > 0:
+        context = data['features'][0].get('context', None)
```
SeanKennyNF, line 47: Context is a misleading name but it comes from the mapbox API so that's why I'm keeping it consistent. This array actually contains information about the parent of the city (it's state, district, country, etc.)

Pathogens/Utility/location_utils/lat_lng_generation_report.py
```diff
@@ -0,0 +1,314 @@
...
+    matching_text_as_string = mapbox_response.matching_text if mapbox_response.matching_text is not None else "N/A"
+    region_name_as_string = mapbox_response.region_name if mapbox_response.region_name is not None else "N/A"
+
+    return "[" + "center_coordinates: " + center_coordinates_as_string \
+        + ", bounding_box: " + bounding_box_as_string \
+        + ", text: " + text_as_string \
+        + ", matching_text: " + matching_text_as_string \
+        + ", region_name: " + region_name_as_string + "]"
+
+def generate_line_for_response_and_request(\
```
SeanKennyNF, line 70: Context: I just want a log for every single request made so I can see how many warnings and errors we get per successful request and also confirm that the cache is still working (when the cache isn't working you get a lot of repeats of the same request over and over)

Pathogens/Utility/location_utils/lat_lng_generation_report.py
```diff
@@ -0,0 +1,314 @@
...
+        logLevel = MapboxApiRequestLogLevel.INFO,
+        city_name = city_name,
+        state_name = state_name,
+        country_name = country_name
+    )
+    request_with_api_key_redacted = "[" + redact_mapbox_api_key_from_mapbox_api_query_url(mapbox_request_url) + "]"
+
+    return prefix + " - " + request_with_api_key_redacted + " - " + format_mapbox_response_as_string(mapbox_response)
+
+def generate_line_for_text_consitency_check(
```
SeanKennyNF, line 87: Context: I want to log a warning every single time the city name we have and the city name mapbox gives are different.

This:
- Caught a lot of typos in our city names (Ex. we have a `Jakarata` instead of `Jakarta` and a `Montsinnery–Tonnegrande` instead of `Montsinéry-Tonnegrande`). It's not perfect and mapbox actually has some typos/questionable names as well but it's at least a flag for the person reading the report to double-check the spelling of the name.
- Is sometimes indicative of a larger issue like mapbox picking the wrong city.

Pathogens/Utility/location_utils/lat_lng_generation_report.py
```diff
@@ -0,0 +1,314 @@
...
+        state_name = state_name,
+        country_name = country_name
+    )
+    info_statement = "[Text returned from the mapbox API did not match expected text]"
+    details = '[Expected text: ' + expected_text_to_display + ', Actual: ' + text_from_mapbox_api_to_display + ']'
+    suggestion = "[If one of the actual text values listed is the correct name of the city/state/country and no other errors or warnings are present, It might be appropriate to change the name of this location in Airtable to the actual name listed. Otherwise, please look to the other warnings and errors for this location for more guidance, the texts not matching might be indicative that mapbox has identified the wrong city based on the text it was given.]"
+
+    return prefix + " - " + info_statement + " - " + details + " - " + suggestion
+
+def generate_line_for_city_state_bounding_box_consistency_check(
```
SeanKennyNF, line 119: Context: This is a big one, basically looks for instances where the city listed doesn't belong to the state listed geographically (Ex. Something like Calgary, Saskatchewan, Canada). Root of a bunch of our wrong locations on the map. Should speak for itself (hopefully?) but hmu if this doesn't make sense.

Pathogens/Utility/location_utils/lat_lng_generation_report.py
```diff
@@ -0,0 +1,314 @@
...
+        state_name = state_name,
+        country_name = country_name
+    )
+    info_statement = "[City listed and state listed are incompatible with one another]"
+    details = '[City name: ' + city_name_as_string + ', State name: ' + state_name_as_string + ']'
+    suggestion = "[According to mapbox, the state this city belongs to is " + region_name_as_string + ". If there are no other errors and you agree with this state placement, consider moving the city to that state. If that state name sounds wrong, please feel free to omit the city name since it might be resulting in the pin being in the incorrect state. Please also check the other errors to see if they provide any more context here.]"
+
+    return prefix + " - " + info_statement + " - " + details + " - " + suggestion
+
+def generate_line_for_invalid_city_but_valid_state_check(
```
SeanKennyNF, line 174: Context: Some of our problems come from someone giving a state name as a city name. Mapbox usually returns `None` in instances like that but sometimes it tries to take an educated guess and it's almost always wrong. Ex `Tyrol, Tyrol, Austria` returns `None` so it falls back to `N/A, Tyrol, Austria` which is valid but in cases like `Madhya Pradesh, Madhya Pradesh, India` it legit just guesses and gets it totally wrong so because this is a mixed bag some of these warnings are actual problems and some of them are just annoying

For what it's worth I think examples like `Tyrol, Tyrol, Austria` should be corrected to `N/A, Tyrol, Austria` but let's merge the report first and then I'll go talk to the data team about how they feel about me making my changes to Airtable.

Pathogens/Utility/location_utils/lat_lng_generation_report.py
```diff
@@ -0,0 +1,314 @@
...
+    )
+
+    info_statement = "[City listed is recognized as both a valid city and a valid state.]" if(mapbox_response is not None) \
+        else "[City listed is not a valid city, but is a valid state]"
+    details = '[City name given: ' + city_name_as_string + ', State name given: ' + state_name_as_string + ']'
+    suggestion = "[According to mapbox, the city you've given is the name of a valid state called " + state_name_from_mapbox_as_string + ". This might be resulting in the pin being in the wrong location because it might be using a city whose name is similar instead of using the state. The city name mapbox lists for this city is " + original_request_text_as_string + " so if that sounds wrong I would recommend moving the current city name to the state column in Airtable if that seems appropriate and makes sense given the other error messages.]"
+
+    return prefix + " - " + info_statement + " - " + details + " - " + suggestion
+
+def generate_line_for_invalid_city_but_valid_district_check(
```
SeanKennyNF, line 216: Context: Same problem as the state check above. Giving a district name as a city name only works if that city is also a district which is usually true but not always and when it's not true Mapbox just takes it's best guess and is almost always in the wrong part of the country.

I really feel like there should be a "District" column in Airtable so that we can get cases like these right but again I've gone super down the rabbit hole so sorry if my rambling doesn't make sense. I'll talk to the data team about it, I just think this report is valuable for illustrating my point.

EliasPinno (approved):

